PR #22, opened 2017-11-16
ci(jenkins): add jenkins distro release job

menski:
*Note*: as the examples are not build during the release the parent version
in the examples pom is not adjusted and has to be modified manually after the
release.

related to zeebe-io/zeebe#549

### DIFF
--- a/.ci/distro_release.dsl
+++ b/.ci/distro_release.dsl
@@ -1,0 +1,202 @@
+// vim: set filetype=groovy:
+
+def jobName = 'spring-zeebe-DISTRO-maven-deploy'
+def repository = 'spring-zeebe'
+def gitBranch = 'master'
+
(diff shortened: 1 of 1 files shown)


menski:
The job only uploads the artifacts to Camunda nexus and maven central. If we want also a GitHub release we need a list for files to upload. Then I can add the GitHub release part to the job.

jangalinski:
The release should contain:

* spring-zeebe-broker
* spring-zeebe-broker-starter

* spring-zeebe
* spring-zeebe-starter

jangalinski:
You are right about the version of the examples ... didn't take that into account. But still I think npot having examples polluting the repo is worth the manual effort

menski:
I added the GitHub release step as an extra commit

PR #55, opened 2019-05-16
Update Zeebe to version 0.17.0

### LINKED ISSUE #50: Version of spring zeebe
Hi all,
Do you foresee to maintain and upgrade spring-zeebe project, I have noticed that the last version us zeebe-client 0.15.0 instead of 0.17.0, the last version of broker.

Thanks a lot,
Matteo.

### LINKED ISSUE #52: support zeebe 0.17.x
current version 0.15.0

menski:
closes #50 
closes #52


jangalinski:
Cool. It seems to be much simplier than expected to switch from payload to variables ... note: since we will be operation with key/value entries now, setting variables on a zeebe process should be as easy as in camunda by `put("key", value)` instead of `set("{\"key\":\"value\"}")` ... 

will review and test tonight!

PR #57, opened 2019-05-16
Rethrow target exception of job worker invocation

### LINKED ISSUE #56: Job Worker exceptions are not propagated
Job workers fail to report exceptions as failed jobs as we do not handle `InvocationTargetExpcetions` in https://github.com/zeebe-io/spring-zeebe/blob/a00daa9e62bccfc796419b0ed59401afd345ee63/client/spring-zeebe/src/main/java/io/zeebe/spring/client/bean/MethodInfo.java#L31

The returned exception will have `null` as exception message which again triggers an exception in the protobuf serializer of the Zeebe client as it is not a valid string value. Also the user does expect the target exception as failed job exception message and not the wrapper.

Reference: https://forum.zeebe.io/t/incident-was-not-created-even-same-service-task-failing-multiple-times/414

menski:
closes #56

### DIFF
--- a/client/spring-zeebe/src/main/java/io/zeebe/spring/client/bean/MethodInfo.java
+++ b/client/spring-zeebe/src/main/java/io/zeebe/spring/client/bean/MethodInfo.java
@@ -3,10 +3,10 @@ package io.zeebe.spring.client.bean;
 import static org.springframework.core.annotation.AnnotationUtils.findAnnotation;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Optional;
 import lombok.Builder;
-import lombok.SneakyThrows;
 import lombok.Value;
 
 @Value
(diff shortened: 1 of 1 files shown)


menski:
I would like to add a test case for this, but I'm not sure how to do this in the current test suites. Any pointers @jangalinski?

jangalinski:
Not sure about the test cases, but I can tell you, that the lombok @SneakyThrows will take every checked exception and wrap it in a RuntimeException, so you possible would want to remove that annotation.
Which will possibly have impact on the lambda-style implementation ...

Could you add a pseudo code test scenario to the issue, then I can think about it ..

menski:
The problem is that any exception throw in a worker implemention would trigger a `InvocationTargetException` with `null` as message. This will lead to a null pointer exception in the Java client as we try to use the exception message as fail job error message. And the user probably wants to see the source error and not the `TargetInvocationException`.

To test this just take any worker implemention of the examples and throw an exception in the implemention.

I think wrapping in a runtime exception is fine as long as it keeps the message and stacktrace. Which seemed to work in my manual tests.

menski:
I removed the `SneakyThrows` and handle exceptions explicitly. Let me know if you have any feedback, otherwise I will merge this. Manuel tests showed the expected behavior :)

jangalinski:
I did not really object, just mentioned that it could be a problem ... if it isn't: perfect! Go ahead

PR #58, opened 2019-05-19
Application.yml correctly used by spring-zeebe-starter

### LINKED ISSUE #49: spring-zeebe-starter is not using the configuration from application.yml
I think there is a problem with spring-zeebe-starter, as it's not working as intended.

It isn't using the configured properties declared in application.yml that maps to io.zeebe.spring.client.properties.ZeebeClientConfigurationProperties.
I think it's just using the defaults configured in: io.zeebe.spring.client.config.ZeebeClientSpringConfiguration#DEFAULT

How to reproduce this problem:

If you start the example (io.zeebe.spring.example.StarterApplication) without having an instance of zeebe's broker running, you'll see this error:

```
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: /0.0.0.0:26500
```

If you change the port in the example's configuration:

```
zeebe:
  broker:
    contactPoint: "localhost:26700"
```

The error keeps showing the default value:

```
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: /0.0.0.0:26500
```

Looking in the structure of ZeebeClientConfigurationProperties, the configuration should be:

```
zeebe:
  client:
    broker:
      contactPoint: "localhost:26700"
```

But that produces the same wrong result as before.

The problem seems to be that the AutoConfiguration (io.zeebe.spring.client.config.ZeebeClientStarterAutoConfiguration) is doing nothing with
the ConfigurationProperties (io.zeebe.spring.client.properties.ZeebeClientConfigurationProperties)

Am I not understanding or using in a bad way the spring-zeebe-starter or there is a problem with the dependency?

Also, I think there is a problem/discrepancy with the worker default name configuration.

io.zeebe.spring.client.annotation.ZeebeWorker declares this:

```
String name() default "${zeebe.worker.name}";
```

But ZeebeClientConfigurationProperties declares this:

```
zeebe:
  client:
    worker:
      name: 
```

Thanks for your great work

spulci:
This hopefully close #49 . 

Please @menski and @jangalinski review carefully because a better way to implement this fix should easily exist (due to my low knowledge of spring).

IMHO what's needed: 

- [x] test to check the yaml has been correctly fetched if present (i.e. default values).

- [x] more tests for other client - worker properties

. Opened to your feedback!

### DIFF
--- a/client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/config/ZeebeClientStarterAutoConfiguration.java
+++ b/client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/config/ZeebeClientStarterAutoConfiguration.java
@@ -4,6 +4,7 @@ import io.zeebe.client.ZeebeClientBuilder;
 import io.zeebe.client.impl.ZeebeClientBuilderImpl;
 import io.zeebe.spring.client.properties.ZeebeClientConfigurationProperties;
 import lombok.RequiredArgsConstructor;
+
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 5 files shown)


spulci:
Tested more client properties and added one more test class to test default value. Does it make sense to test defaults even if they could change in a near future?

PR #74, opened 2019-11-24
Deployment of multiple resources by annotation

THRuM:
Hi
I found possibility of deployment only one resource using `ZeebeDeployment` annotation very limiting, not much to modify and it works with multiple resources. Could be useful also for others, let me know what you think.


salaboy:
@THRuM this is a great contribution.. can you add a test with multiple classpath resources? 
to make sure that it will not fail

THRuM:
@salaboy I've added unit tests, I wanted to add also integration test but I am not shure how to approach to them. I wanted to use `ZeebeTestRule` alongsite with `ZeebeDeployment` but it seems to be more tricky :smile:. Is there an example of `ZeebeDeployment` used in integration tests somewhere?

salaboy:
@THRuM I was looking at the tests and those are good enough for me right now.. I am planning for more integration tests but at the Kubernetes Level.. which might not be for everyone..

PR #87, opened 2020-04-07
zeebe client health indicator

### LINKED ISSUE #79: Add HealthIndicator to add basic ZeebeConnection info available via Actuator
Add a Spring Boot Actuator (https://www.baeldung.com/spring-boot-actuators) Helth indicator to the Zeebe client, that will switch to unhealthy if the client connection is not successful.

A TypeScript example using the NodeJS client library did this by using GRPC hooks (https://github.com/camunda-cloud/zeebe-http-worker/blob/master/src/worker/worker.ts#L86) - probably possible in Java as well?

pabloromanelli:
Zeebe Health Indicator using Spring Actuator as optional dependency.
Closes #79

### DIFF
--- a/client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/actuator/ZeebeClientHealthIndicator.java
+++ b/client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/actuator/ZeebeClientHealthIndicator.java
@@ -1,0 +1,26 @@
+package io.zeebe.spring.client.actuator;
+
+import io.zeebe.client.ZeebeClient;
+import io.zeebe.client.api.response.Topology;
+import org.springframework.boot.actuate.health.AbstractHealthIndicator;
+import org.springframework.boot.actuate.health.Health;
+
+public class ZeebeClientHealthIndicator extends AbstractHealthIndicator {
+
+  private final ZeebeClient client;
+
+  public ZeebeClientHealthIndicator(ZeebeClient client) {
+    this.client = client;
+  }
+
+  @Override
+  protected void doHealthCheck(Health.Builder builder) {
+    Topology topology = client.newTopologyRequest().send().join();
+    if (topology.getBrokers().isEmpty()) {
+      builder.down();
+    } else {
+      builder.up();
+    }
+  }
+
+}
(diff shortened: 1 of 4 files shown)


client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/actuator/ZeebeClientHealthIndicator.java
```diff
@@ -0,0 +1,26 @@
...
+public class ZeebeClientHealthIndicator extends AbstractHealthIndicator {
+
+  private final ZeebeClient client;
+
+  public ZeebeClientHealthIndicator(ZeebeClient client) {
+    this.client = client;
+  }
+
+  @Override
+  protected void doHealthCheck(Health.Builder builder) {
```
berndruecker, line 17: Hey @pabloromanelli - thank you so much for this PR! Happy to merge it before I release the 0.23 for Spring :-) One quick question: Do you know what happens if the broker there is an exception raised in this method (as if the broker is really down)? does this count as unhealthy?

pabloromanelli: Happy to help :)
It should count as unhealthy because in `AbstractHealthIndicator` they catch exceptions thrown by `doHealthCheck` to produce a "down" status:

```java
	public final Health health() {
		Health.Builder builder = new Health.Builder();
		try {
			doHealthCheck(builder);
		}
		catch (Exception ex) {
			if (this.logger.isWarnEnabled()) {
				String message = this.healthCheckFailedMessage.apply(ex);
				this.logger.warn(StringUtils.hasText(message) ? message : DEFAULT_MESSAGE, ex);
			}
			builder.down(ex);
		}
		return builder.build();
	}
```

berndruecker: How it should and how I expected it. Great! Thanks!

salaboy:
@berndruecker you merged this.. but this actually changed the behaviour of the Zeebe Spring extension.. now the whole application will be marked as DOWN if the Zeebe Cluster cannot be reached.. I wouldn't do that by default..

berndruecker:
I see what you mean. For most workers I think this is a useful default behavior - but I agree that it might be a very harsh thing to do in the general spring client.

I would love to see some docs on this flag in the readme, then I am happy to change the default. Can you probably push that?

PR #91, opened 2020-04-20
fix(client): use default values for job workers from config

### LINKED ISSUE #88: The job timeout is not applied from the config
## Problem

When I configure the timeout of a job in the configuration then it is not applied to the worker that is annotated with `@ZeebeWorker`.

```
zeebe.client.job.timeout=10s
```

It works only if I define the timeout in the annotation itself.

## Solution

* the job timeout is applied from the config

### LINKED ISSUE #89: The maxJobsActive value is not applied from the config
## Problem

When I configure the `maxJobsActive` of a job in the configuration then it is not applied to the worker that is annotated with `@ZeebeWorker`.

```
zeebe.client.worker.maxJobsActive=10
```

It works only if I define the timeout in the annotation itself.

## Solution

* the worker `maxJobsActive` is applied from the config

saig0:
* use job timeout from config if it is not defined by @ZeebeWorker
* use worker maxJobsActive from config if it is not defined by @ZeebeWorker

closes #88 
closes #89

### DIFF
--- a/client/spring-zeebe/src/main/java/io/zeebe/spring/client/annotation/ZeebeWorker.java
+++ b/client/spring-zeebe/src/main/java/io/zeebe/spring/client/annotation/ZeebeWorker.java
@@ -15,7 +15,7 @@ public @interface ZeebeWorker {
 
   String name() default "${zeebe.worker.defaultName}";
 
-  long timeout() default 10000L;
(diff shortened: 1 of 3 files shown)


berndruecker:
Great! Thanks.

PR #93, opened 2020-04-23
feat(cloud): Easy configuration for Camunda Cloud (related to https:/…

berndruecker:
…/github.com/zeebe-io/spring-zeebe/issues/81)


berndruecker:
@saig0 @ArtunSubasi @salaboy Ok with you? Or comments?

saig0:
Nice :+1: 
From the README, it is not clear that the `credentialsCachePath` is optional but it seems so.

berndruecker:
You are right - I simply removed it to avoid that people need to think (why should I adjust it anyway?)

ArtunSubasi:
> You are right - I simply removed it to avoid that people need to think

Bernd, I like the way you think 👍

PR #101, opened 2020-08-08
configure generic oauth in zeebe.client.security.oauth separate from zeebe.client.cloud

cameronbraid:
I would like to use oauth config where I can customize the audience field.  Currently this is generated from the cloud config.  

Two possible solutions come to mind 
* add an `String audience` field into cloud
* add a separate oauth config block outside cloud

I chose the second option as this is conceptually not a cloud configuration

So you can do the following : in application.yaml

```
zeebe:
    client:
      worker:
        defaultName: worker-name
      broker.contactPoint: 127.0.0.1:26500
      security:
            plaintext: true
            oauth:
                  clientId: xxx
                  clientSecret: xxx
                  authUrl: xxx
                  audience: xxx
                  credentialsCachePath: xxx
```

### DIFF
--- a/client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
+++ b/client/spring-zeebe-starter/src/main/java/io/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
@@ -71,6 +71,15 @@ public class ZeebeClientConfigurationProperties implements ZeebeClientProperties
 	}
   }  
 
+
+  @Data
+  public static class OAuth {
+    private String clientId;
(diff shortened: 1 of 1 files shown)


salaboy:
@cameronbraid are you still working on this? Can you please update the status of this PR?

berndruecker:
Hi @cameronbraid - thanks for the PR! I currently try to understand your use case for customizing the audience field - could you elaborate? I would like to either move forward with this or close the PR - to get spring zeebe in shape for 1.0 :-)

berndruecker:
Closing this for now as I haven't heard back. Let me know if there is still something you need!

PR #113, opened 2021-03-24
Adds E2E example test for WorkerApplication

npepinpe:
**Description**

This PR adds an example test in the worker example for an E2E test using Testcontainers. It will spin up a Zeebe broker with standalone gateway and wait until it's ready. The worker application is then started and autowired to this container, will deploy its process, and start its workers.

A simple test of creating an instance and waiting for its result is added, which can be expanded to add more complicated tests.

It could be interesting to also add later an example with zeeqs such that users can do more complex verifications (e.g. are my processes deployed? how many processes are running? etc.).

Some notes:

- I wasn't where was the best place to put this, feel free to move this around
- I initially used the configuration as described in the readme, `zeebe.client.broker.gatewayAddress`, but this didn't work. Changing it to the deprecated `zeebe.client.broker.contactPoint` worked however.
- Adding this means adding a development pre-requisite of Docker - to run the test you will need a Docker environment as described [here](https://www.testcontainers.org/supported_docker_environment/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: java
 sudo: false # faster builds
 
+services:
+  - docker
+
 jdk:
 - openjdk8
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ pipeline {
       cloud getZeebeK8sCloud()
       label getZeebeK8sLabel()
       defaultContainer 'jnlp'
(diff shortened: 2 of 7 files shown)


npepinpe:
Small note: 2.0.0 was released very recently, so it seems it's still not available on Maven Central. Should be there in a second though, and Travis builds should also pass I hope.

PR #114, opened 2021-04-13
#109: Add an example of using a custom `JsonMapper`

### LINKED ISSUE #109: Ability override ZeebeObjectMapper
Hi, I want to override ZeebeObjectMapper or customize it.
I use Kotlin and default objectMapper`s settings don`t support kotlin module.

I use a workaround - deserialize of variables with my objectMapper, but that`s not what I want

aivinog1:
closes #109


berndruecker:
Awesome - thx

PR #118, opened 2021-06-02
support configure resources with classpath:*

Lzgabel:
* feat(spring-zeebe): support configure resources with classpath:*
* be compatible with "classPathResources" and mark it as deprecated

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ You can also deploy multiple files at once:
 
 ```
 @ZeebeDeployment(classPathResources ={"demoProcess.bpmn", "demoProcess2.bpmn"})
+@ZeebeDeployment(resources = "classpath*:/bpmn/**/*.bpmn")
 ```
 
 ## Implement Job Worker
--- a/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/value/ZeebeDeploymentValue.java
+++ b/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/value/ZeebeDeploymentValue.java
@@ -6,17 +6,17 @@ import java.util.Objects;
 
 public class ZeebeDeploymentValue implements ZeebeAnnotationValue<ClassInfo> {
 
-  private List<String> classPathResources;
+  private List<String> resources;
 
   private ClassInfo beanInfo;
 
-  private ZeebeDeploymentValue(List<String> classPathResources, ClassInfo beanInfo) {
-    this.classPathResources = classPathResources;
+  private ZeebeDeploymentValue(List<String> resources, ClassInfo beanInfo) {
+    this.resources = resources;
     this.beanInfo = beanInfo;
   }
 
-  public List<String> getClassPathResources() {
-    return classPathResources;
+  public List<String> getResources() {
+    return resources;
   }
 
   @Override
@@ -29,19 +29,19 @@ public class ZeebeDeploymentValue implements ZeebeAnnotationValue<ClassInfo> {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     ZeebeDeploymentValue that = (ZeebeDeploymentValue) o;
-    return Objects.equals(classPathResources, that.classPathResources) &&
+    return Objects.equals(resources, that.resources) &&
       Objects.equals(beanInfo, that.beanInfo);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(classPathResources, beanInfo);
+    return Objects.hash(resources, beanInfo);
   }
 
   @Override
   public String toString() {
     return "ZeebeDeploymentValue{" +
-      "classPathResources=" + classPathResources +
+      "resources=" + resources +
       ", beanInfo=" + beanInfo +
       '}';
   }
@@ -52,14 +52,14 @@ public class ZeebeDeploymentValue implements ZeebeAnnotationValue<ClassInfo> {
 
   public static final class ZeebeDeploymentValueBuilder {
 
-    private List<String> classPathResources;
+    private List<String> resources;
     private ClassInfo beanInfo;
 
     private ZeebeDeploymentValueBuilder() {
     }
 
-    public ZeebeDeploymentValueBuilder classPathResources(List<String> classPathResources) {
-      this.classPathResources = classPathResources;
+    public ZeebeDeploymentValueBuilder resources(List<String> resources) {
+      this.resources = resources;
       return this;
     }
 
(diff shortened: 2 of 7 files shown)


Lzgabel:
Hi, @jwulf   Can you review it? :)

jwulf:
> Hi @jwulf, Can you review it? :)

I'm not the best person @Lzgabel  - I've requested review from Jan, Bernd, or Sebastian.

Lzgabel:
> I'm not the best person @Lzgabel - I've requested review from Jan, Bernd, or Sebastian.

Thanks :)

berndruecker:
Hey @Lzgabel - I quickly looked at it and try to understand the why of this change. Is the intention to be more flexible on BPMN locations? 
Shouldn't this then deprecate the "classPathResources" options - as it is wired to have both options - WDYT?

@jangalinski Do you have by chance a spontaneous opinion on this? I think you are the bets person to comment on it :-)

jangalinski:
I will look into it, hopefully til the end of the week

Lzgabel:
> Is the intention to be more flexible on BPMN locations?

Yes, in the actual use scenario, we have many bpmn files in different folders to deploy.

> Shouldn't this then deprecate the "classPathResources" options - as it is wired to have both options - WDYT?

I agree with this point, I will mark "classPathResources" deprecated.

berndruecker:
I would love to still get Jans feedback - but I do see the advantage of having the full Spring flexibility to reference resources - so I am supportive of this change.

However, the attribute "bpmnLocations" should not be named like this, as you might be able  to deploy different things, at least down the road (e.g. forms). And with "@ZeebeDeployment" the context is clear enough, hence I would simply name it "resources=...". WDYT?

And when deprecating the "classPathResources" - could we maybe do it in such a way, that it is simply used as "resources = classpath:VALUE_FROM_classPathResources" - so that the whole handling of the resources is identical and we avoid any differences in behavior?

Hope that make sense?

README.md
```diff
@@ -44,6 +44,7 @@ You can also deploy multiple files at once:
 
 ```
 @ZeebeDeployment(classPathResources ={"demoProcess.bpmn", "demoProcess2.bpmn"})
+@ZeebeDeployment(bpmnLocations = "classpath*:/bpmn/**/*.bpmn")
```
berndruecker, line 47: "resources" instead of "bpmnLocations"

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/value/ZeebeDeploymentValue.java
```diff
@@ -8,11 +8,14 @@
 
   private List<String> classPathResources;
 
+  private List<String> bpmnLocations;
+
   private ClassInfo beanInfo;
 
-  private ZeebeDeploymentValue(List<String> classPathResources, ClassInfo beanInfo) {
+  private ZeebeDeploymentValue(List<String> classPathResources, ClassInfo beanInfo, List<String> bpmnLocations) {
```
berndruecker, line 15: I think it should be either/or - not both attributes. And "classPathResources" could be translated into "resources" as I tried to mention in the overall comment

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/value/ZeebeDeploymentValue.java
```diff
@@ -8,11 +8,14 @@
 
   private List<String> classPathResources;
 
+  private List<String> bpmnLocations;
```
berndruecker, line 11: I don#t think we need two attributed - "resources" should be sufficient - and the old "classPathResources" can be translated into the right Spring resources string

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/config/processor/DeploymentPostProcessor.java
```diff
@@ -40,9 +53,25 @@ public boolean test(final ClassInfo beanInfo) {
...
+
+      Stream<DeployProcessCommandStep1.DeployProcessCommandBuilderStep2> deployProcessCommandBuilderStep2StreamOfLocations = locations.stream().map(resource -> {
+        try {
+          return deployWorkflowCommand.addResourceStream(resource.getInputStream(), resource.getFilename());
+        } catch (IOException e) {
+        }
+        return null;
+      }).filter(Objects::nonNull);
+
+      Stream<DeployProcessCommandStep1.DeployProcessCommandBuilderStep2> deployProcessCommandBuilderStep2StreamOfResources = value.getClassPathResources()
```
berndruecker, line 69: If we condense everything into one attribute - we don't need duplicated code here :-)

client/spring-zeebe/src/test/java/io/camunda/zeebe/spring/client/bean/value/factory/ReadZeebeDeploymentValueTest.java
```diff
@@ -42,6 +42,7 @@ public void shouldReadSingleClassPathResourceTest() {
     ZeebeDeploymentValue expectedDeploymentValue = ZeebeDeploymentValue.builder()
       .beanInfo(classInfo)
       .classPathResources(Collections.singletonList("/1.bpmn"))
+      .bpmnLocations(Collections.singletonList("classpath*:/bpmn/**/*.bpmn"))
```
berndruecker, line 45: I would prefer to have two test cases - one testing the legacy "classPathResources" attribute - and one the new "resources" attribute. I don't think it is valid to use both at the same place - maybe we should even disallow this.

berndruecker (commented):
I added some comments - hope that all make sense for you. Thanks for all your work on this!!

Lzgabel:
> I added some comments - hope that all make sense for you. Thanks for all your work on this!!

Hi @berndruecker, thanks for the quick reply, it's done, Can you review it? :)

README.md
Lzgabel: I have renamed "bpmnLocations" to "resources"

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/value/ZeebeDeploymentValue.java
Lzgabel: I see your point.

berndruecker:
Fixes https://github.com/zeebe-io/spring-zeebe/issues/120

berndruecker:
Looks good to me - let me merge it and have another quick look next week on master

berndruecker:
Thank you @Lzgabel!

berndruecker:
I updated the readme @Lzgabel: https://github.com/zeebe-io/spring-zeebe/commit/1df0ce9d59426dd7004832929f4ab13bf591aee6#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5
Could you have a quick look?

We will release 1.1.0 later today and this change will be in it. Thx again!

PR #124, opened 2021-08-19
Suggestions about @Repository annotation usage

WilliamSESE:
Hi, I found that there may be some minor improvements about annotations in your code.

It would be more appropriate to add @repository annotation at the class level. The @repository annotation indicates that the class provides the mechanism for storage, retrieval, search, update and delete operation on objects.

For better understanding and maintainability of a large application, adding @repository is a better choice.

Hope my suggestions are useful!


jangalinski (changes requested):
I chose to not annotate components and configurations, so they never are accidentally scanned and added to the context. Everything is explicit, meaning you either import the configuration manually or use the "Enable" annotation (which just automates the import).

I would stick to this for 3rd party libs like spring-zeebe.
That being said, I would not like this to be merged and it is not required to run the spring client.

Could you provide a (failing) use case that motivated you to make this change?

Issue #130, opened 2021-10-04
Spring Boot properties using camel case in README are outdated


rob2universe:
should be:
```
zeebe.client.cloud.cluster-id
zeebe.client.cloud.client-secret
Zeebe.client.cloud.client-id
zeebe.client.cloud.base-url
zeebe.client.cloud.auth-url
zeebe.client.worker.max-jobs-active
```

instead of 
```
zeebe.client.cloud.clusterId
zeebe.client.cloud.clientSecret
zeebe.client.cloud.clientId
zeebe.client.cloud.baseUrl
zeebe.client.cloud.authUrl
zeebe.client.worker.maxJobsActive
```

berndruecker:
Hey @rob2universe - why is camelCase outdated? pring still does support that - or not?

berndruecker:
OK - sounds reasonable to me. Are you willing to provide a PR for this? It is Hacktoberfest after all :)
https://camunda.com/hacktoberfest2021/

berndruecker:
Closed by https://github.com/camunda-community-hub/spring-zeebe/pull/131

PR #131, opened 2021-10-22
chore #130 update Spring Boot properties in README

rob2universe:


berndruecker:
Thanks a lot @rob2universe

PR #135, opened 2021-11-11
Support fetch of job variables from param's annotation

### LINKED ISSUE #132: Make fetching of Job variables more convenient and DRY
Instead of repeating variable names in the `fetchVariables` property of the `@ZeebeWorker` annotation and then later in the actual variable access, e.g.

```java
  @ZeebeWorker(type = "myType", fetchVariables = { "orderId" } )
  public void handleJob(final JobClient client, final ActivatedJob job) {
    String orderNumber = (String) job.getVariablesAsMap().get("orderId");
  }
```

we could keep it DRY (Don't Repeat Yourself) and convenient with an argument list, e.g.

```java
  @ZeebeWorker(type = "myType")
  public void handleJob(final JobClient client, final ActivatedJob job, String orderId) {
  }
```

Type casting would also be done implicitly here. Not sure if the additional parameter would have to be annotated to make this work.

Lzgabel:
* support fetch of job variables from param's annotation

closes #132

### DIFF
--- a/client/spring-zeebe/src/test/java/io/camunda/zeebe/spring/client/bean/MethodInfoTest.java
+++ b/client/spring-zeebe/src/test/java/io/camunda/zeebe/spring/client/bean/MethodInfoTest.java
@@ -1,0 +1,68 @@
+package io.camunda.zeebe.spring.client.bean;
+
+import io.camunda.zeebe.spring.client.annotation.ZeebeVariable;
+import io.camunda.zeebe.spring.client.annotation.ZeebeWorker;
+import org.junit.Test;
+import org.mockito.internal.util.collections.Sets;
+
+import java.beans.Introspector;
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class MethodInfoTest {
+
+  public static class WithZeebeWorker {
+
+    @ZeebeWorker(type = "bar", timeout = 100L, name = "kermit")
+    public void handle() {
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


client/spring-zeebe/src/test/java/io/camunda/zeebe/spring/client/bean/MethodInfoTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @ZeebeWorker(type = "bar", timeout = 100L, name = "kermit")
+    public void handle() {
+    }
+  }
+
+  public static class WithZeebeWorkerParamValues {
+
+    @ZeebeWorker(type = "foo", name = "kermit")
+    public void handle(@ZeebeVariable("foo") String foo) {
```
falko, line 30: Is it necessary to explicitly define the variable name or could we just use the argument name?

Lzgabel: @falko thank you so fast replay!

Yes, you can't get the real variable name when the code is compiled

berndruecker:
Hey @Lzgabel - thanks! I already started a stab at it with some differences. I want to look into this in more detail but might not come to it this week. Sorry for the delay!

berndruecker:
Hey @Lzgabel - I wanted to let you know ASAP that I had another look and would close this PR in favor of the code I have locally. Sorry for the overlap! I could draw some inspiration out of this code though :-) I know mark the issue as in progress correctly and can hopefully push my code to a branch soon. Just discovered that the real invocation is actually not tested by any test case 😨

Issue #143, opened 2021-11-16
Allow auto-complete of jobs


berndruecker:
Within a job handler, everybody always has to complete the job themselves. Doing this right is not 100% obvious to beginners (e.g. doing asynchronous completion, see https://blog.bernd-ruecker.com/writing-good-workers-for-camunda-cloud-61d322cad862).

So instead of writing:

```
@ZeebeWorker(type = "foo")
public void handleJobFoo(final JobClient client, final ActivatedJob job) {
  // do whatever you need to do
  client.newCompleteCommand(job.getKey()) 
     .variables("{\"fooResult\": 1}")
     .send()
     .exceptionally( throwable -> { throw new RuntimeException("Could not complete job " + job, throwable); });
}
```

one can write:

```
@ZeebeWorker(type = "foo", autoComplete = true)
public void handleJobFoo(final JobClient client, final ActivatedJob job) {
  // do whatever you need to do
}
```

Now there are the following features one can leverage:

1. returning a map means: These are variables to add to the process
2. throwing a `BpmnError` means, that this error should be raised via API
3. throwing any other exception means, that this exception is rethrown, leading to retries like normal Java Client behavior (https://github.com/camunda-cloud/zeebe/blob/develop/clients/java/src/main/java/io/camunda/zeebe/client/impl/worker/JobRunnableFactory.java#L45)

berndruecker:
@falko Any comments?

berndruecker:
Added a first stab at the code on the branch, documented here: https://github.com/camunda-community-hub/spring-zeebe/tree/worker-convenience#completing-the-job

berndruecker:
Closed by https://github.com/camunda-community-hub/spring-zeebe/pull/144

PR #144, opened 2021-11-17
Adding convenience (variable fetch, auto complete) for workers

berndruecker:
Fixes https://github.com/camunda-community-hub/spring-zeebe/issues/132 and https://github.com/camunda-community-hub/spring-zeebe/issues/143


berndruecker:
@falko Could you probably also have a look?

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/value/ZeebeWorkerValue.java
```diff
@@ -16,19 +24,32 @@
...
     this.beanInfo = beanInfo;
+
+    if (forceFetchAllVariables) {
+      // this overwrites any other setting
+      this.fetchVariables = new String[0];
+    } else {
+      // make sure variables configured and annotated parameters are both fetched, use a set to avoid duplicates
+      Set<String> variables = new HashSet<>();
+      variables.addAll(Arrays.asList(fetchVariables));
+      variables.addAll(variableParameters.stream().map(p -> p.getParameterName()).collect(Collectors.toList()));
```
menski, line 50:
```suggestion
      variables.addAll(variableParameters.stream().map(ParameterInfo::getParameterName).collect(Collectors.toList()));
```

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/config/processor/ZeebeWorkerSpringJobHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private static final Logger LOG = Loggers.JOB_WORKER_LOGGER;
+  private ZeebeWorkerValue workerValue;
+
+  public ZeebeWorkerSpringJobHandler(ZeebeWorkerValue workerValue) {
+    this.workerValue = workerValue;
+  }
+
+  @Override
+  public void handle(JobClient jobClient, ActivatedJob job) throws Exception {
```
menski, line 31: I have the feeling that a lot of this figuring out the parameter classes and result class can be done once, and should be done on every new job. But I think it's something we can improve afterwards and doesn't have to block this improvement.

menski (changes requested):
I really like this feature, I didn't used it yet, but it looks quite convenient

README.md
```diff
@@ -76,6 +76,103 @@ public void handleJobFoo(final JobClient client, final ActivatedJob job) {
...
 
+### Fetch Variables
+
+You can access all variables of a process via the job:
+
+```
+@ZeebeWorker(type = "foo")
+public void handleJobFoo(final JobClient client, final ActivatedJob job) {
+  String variable1 = (String)job.getVariablesAsMap().get("variable1");
+  sysout(variable1);
```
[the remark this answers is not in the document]
berndruecker: It is not a thing, this way this would require a method with this name. I doubt that code will really be used - but I am also happy to change it to System.outprintln 👍

README.md
```diff
@@ -76,6 +76,103 @@ public void handleJobFoo(final JobClient client, final ActivatedJob job) {
...
+```
+
+When using `autoComplete` you can:
+
+* Return a `Map`, `String`, `InputStream`, or `Object`, which then will be added to the process variables
+* Throw a `ZeebeBpmnError` which results in a BPMN error being sent to Zeebe
+
+```
+@ZeebeWorker(type = "foo", autoComplete = true)
+public void handleJobFoo(final JobClient client, final ActivatedJob job) {
```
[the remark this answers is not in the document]
berndruecker: Good catch 👍

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/config/processor/ZeebeWorkerSpringJobHandler.java
berndruecker: I agree with both comments :-)

berndruecker: I add a comment to the existing code

berndruecker:
Reviewed with @falko today

PR #178, opened 2022-02-11
Bugfix: This could not be interpreted by spring

jonathanlukas:
When trying to use the parameter from the `application.yaml`, the worker could not be started.

Convention is to use `-` instead of camelCase for `application.yaml` files. Usually, spring can resolve this when mapping yaml values to Java classes, but it seems that for an expression, the `-` way is required.


berndruecker:
Awesome - thx!

PR #182, opened 2022-03-16
Recommend direct dependency to latest Zeebe client

falko:
Although Spring Zeebe has a transitive dependency to the [Zeebe Java Client](https://docs.camunda.io/docs/apis-clients/java-client/), we recommend adding a direct dependency to it's latest version in your `pom.xml` as the release of Spring Zeebe is currently not linked to the Zeebe release cycle

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,15 @@ Add the following Maven dependency to your Spring Boot Starter project:
 </dependency>
 ```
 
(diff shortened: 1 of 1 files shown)


berndruecker:
I wonder if instead, we should try to release spring-zeebe often enough so that this should not be necessary (even if possible in certain scenarios?)

berndruecker:
I added a slightly weaker version in the readme that get's pushed with the C8 changes

PR #190, opened 2022-04-15
Deploy DMN decision models via new deploy resource rpc

Lzgabel:
The Java client was already able to deploy DMN decision models, deploy DMN decision models via new deploy resource rpc.

### DIFF
--- a/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/config/processor/DeploymentPostProcessor.java
+++ b/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/config/processor/DeploymentPostProcessor.java
@@ -2,25 +2,30 @@ package io.camunda.zeebe.spring.client.config.processor;
 
 import io.camunda.zeebe.client.ZeebeClient;
 import io.camunda.zeebe.client.api.command.DeployProcessCommandStep1;
+import io.camunda.zeebe.client.api.command.DeployResourceCommandStep1;
+import io.camunda.zeebe.client.api.response.Decision;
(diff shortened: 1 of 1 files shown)


Lzgabel:
Hi @berndruecker. Please check this out in your free time. 🙇

berndruecker:
Code looks good to me! 
But please note, that the existing code is also able to deploy DMN already (the deployProcess also takes DMN :-)).

My only concern I have is that this breaks backward compatibility to pre-8.0 versions (for a relatively small gain). @falko Do you have a spontaneous thought on this matter?

falko:
I found it confusing that the old command does not list the DMN files. It made me think that it didn't work and I did something wrong.

Could we check with reflection if the new command is there and use it if it's there? The old one could be used as a fallback maybe combined with a warning that an old Zeebe client is used.

A warning that the Zeebe client version doesn't match the engine version could be a nice feature for Spring Zeebe anyway.

falko:
@nevries feel free to chime in here.

nevries:
I just hit the same issue, that is, DMN deployment works but the output does imply it wouldn't. I don't see a big issue with loosing compatibility with older client versions as the versioning of spring-zeebe was also updated to 8.x.x so having a client with this major release number makes sense to me. If there is a strong argument against it, I would support Falko's suggestion (reflection) given that the new command would fix the output problem.

berndruecker:
Thanks for the input folks!

While I like the idea of reflection and the warning about versions, I don't see this as the biggest priority at the moment given the lack of time. 
I agree that breaking backward compatibility is not a big deal given we switched to 8.x - so I will merge this now.

PR #195, opened 2022-04-27
Add JsonMapper

falko:
Provides the `JsonMapper` of the `ZeebeClient` as a Spring Bean that can be injected, e.g. for logging or debugging variable objects.

Usage example:

```
@Autowired
private JsonMapper mapper;
 
@ZeebeWorker(type = "my-service")
public void invokeMyService(@ZeebeVariablesAsType ProcessVariables variables) {
  LOG.info("Invoking myService with variables: " + mapper.toJson(variables));
}
```

A unit test is included.


falko:
Switching to draft because I found an issue with a cyclic dependency that I'm currently investigating

falko:
The conflict is with #195/#109

chDame:
Hi Falko,

as discussed, I think that it can be pretty handy for backend developers to inject their own JsonMapper to change its behaviour. In most of the case, the default behavior will be the desired one. But in some project, developers could want to add some extra fields in the generated Json (classname for example) or change the attribute naming style.

PR #209, opened 2022-06-06
fix: remove the duplicated worker name initializations

### LINKED ISSUE #204: The default worker name property doesn't work
So, I think the problem in this code:
https://github.com/camunda-community-hub/spring-zeebe/blob/master/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/processor/ZeebeWorkerAnnotationProcessor.java#L81-L86
If the value isn't present we are setting the value from the class name and method name and not trying to set it from 
the `zeebe.client.worker.defaultName` property.

aivinog1:
I think that we should rely in this case only on `io.camunda.zeebe.spring.client.annotation.processor.ZeebeWorkerAnnotationProcessor.start` as a single source of truth.
Closes #204


berndruecker:
Thank @aivinog1 for looking into this - you are absolutely right (and I think we should probably also add a test). I was out last week - hence the delay...

berndruecker:
Thank @aivinog1 for looking into this - you are absolutely right (and I think we should probably also add a test). I was out last week - hence the delay...

PR #221, opened 2022-08-24
feat: Add ability to disable workers.

### LINKED ISSUE #220: I as a Zeebe worker's developer want to be able to disable workers via configuration properties
We have such a case: we are splitting our big application with a large number of workers into smaller more domain-like applications. So it would be so handy if we could disable a worker in our old application to test the worker only on the new one. To simplify this process I think that would be nice to have the ability to disable workers via configuration properties.

I've played with code a little and I will provide a PR soon :)

aivinog1:
So, it could be disabled in the `ZeebeWorker` annotation and via configuration properties.

Closed #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,16 +384,7 @@ class MyConfiguration {
 
 ### Disable worker
 
-If you need to disable worker you can do it using properties:
-```yaml
-zeebe:
-  client:
-    worker:
-      workers:
-        someWorker:
-          enabled: false
-```
-Or using the `enabled` `ZeebeWorker`'s annotation parameter:
+There exists an ability to disable a worker via the `enabled` `ZeebeWorker`'s annotation parameter:
 ```java
 import io.camunda.zeebe.spring.client.annotation.ZeebeWorker;
 
@@ -404,6 +395,22 @@ class SomeClass {
   }
 }
 ```
+It could be useful when there is a migration of a worker between different applications. This process is described below.
+
+### Override values in the `io.camunda.zeebe.spring.client.annotation.ZeebeWorker` annotation
+
+Sometimes it is helpful to override the `ZeebeWorker` annotation's values.
+For example, if there are two applications and there is a migration of workers from one to another it is convenient to disable workers via configuration in the old application.
+This could be done like this:
+```properties
+zeebe.client.worker.override.my-worker.enabled=false
+```
+where `my-worker` is the type of the worker that we want to customize.
+Also, there all configuration supported that exists in the `ZeebeWorker` annotation:
+```properties
+zeebe.client.worker.override.my-worker.timeout=10000
+```
+It is also possible to provide a different way to customize the `ZeebeWorker` annotation by implementing the `io.camunda.zeebe.spring.client.annotation.customizer.ZeebeWorkerValueCustomizer` interface.
 
 # Code of Conduct
 
--- a/README.md
+++ b/README.md
@@ -382,6 +382,28 @@ class MyConfiguration {
 }
 ```
 
+### Disable worker
+
+If you need to disable worker you can do it using properties:
+```yaml
+zeebe:
+  client:
+    worker:
+      workers:
+        someWorker:
+          enabled: false
+```
+Or using the `enabled` `ZeebeWorker`'s annotation parameter:
+```java
+import io.camunda.zeebe.spring.client.annotation.ZeebeWorker;
+
+class SomeClass {
+  @ZeebeWorker(type = "foo", enabled = false)
+  public void disabledHandleJobFoo(final JobClient client, final ActivatedJob job) {
+    // worker's code
+  }
+}
+```
 
 # Code of Conduct
 
--- a/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
+++ b/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
@@ -1,6 +1,7 @@
 package io.camunda.zeebe.spring.client.properties;
 
 import io.camunda.zeebe.client.impl.ZeebeClientBuilderImpl;
+import io.camunda.zeebe.spring.client.annotation.value.ZeebeWorkerValue;
 import io.grpc.ClientInterceptor;
 import io.camunda.zeebe.client.CredentialsProvider;
 import java.time.Duration;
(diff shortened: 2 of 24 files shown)


aivinog1:
Hey @berndruecker! It is ready for your review :) 

So, I add the CI pipeline on the PRs, if you don't mind. I will be appreciated it if you will take a look when you have some time :) Thank you in advance :)

aivinog1:
Hey @berndruecker! This is a kind reminder about this PR :) Can you check this out, please? :)

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: java
 sudo: false # faster builds
 
 jdk:
-- openjdk8
+- openjdk17
```
jonathanlukas, line 5: Why would you upgrade here? Java 1.8 is still supported as LTS and as this is a framework, we should try to stick to the bottom line of compatibility.

aivinog1: Thanks for pointing this out 👍 But I stick with [GitHub actions](https://github.com/camunda-community-hub/spring-zeebe/blob/master/.github/workflows/mvn-build.yml#L20) and I see here the java requirement for the 17th version: https://github.com/camunda-community-hub/spring-zeebe/blob/568d70df542202c35d4a89f49c9a7a1bde230d08/test/common/pom.xml#L16

aivinog1: I'm not sure about Travis in this project at all 😅 Since there are already Github actions :)

jonathanlukas: Awesome. This requires some additional cleaning I guess...

https://github.com/camunda-community-hub/spring-zeebe/issues/223

jonathanlukas (approved):

berndruecker:
Sorry - vacation time. I try to look at it asap

README.md
```diff
@@ -382,6 +382,28 @@ class MyConfiguration {
...
+### Disable worker
+
+If you need to disable worker you can do it using properties:
+```yaml
+zeebe:
+  client:
+    worker:
+      workers:
+        someWorker:
+          enabled: false
```
berndruecker, line 394: First: Other parts of the doc use properties - not YAML, so I would like to keep it consistent. Like this - right?

zeebe.client.worker.workers[0].someWorker.enabled=false

Then this uses the 'type' as identifier - right? So sticking to the exymplw code below it would be

zeebe.client.worker.workers[0].foo.enabled=false
or
zeebe.client.worker.workers['foo'].enabled=false
?

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/value/factory/ReadAnnotationValueConfiguration.java
```diff
@@ -19,8 +20,8 @@ public ReadZeebeDeploymentValue readZeebeDeploymentValue(final ZeebeExpressionRe
   }
 
   @Bean
-  public ReadZeebeWorkerValue readZeebeWorkerValue(final ZeebeExpressionResolver resolver) {
-    return new ReadZeebeWorkerValue(resolver);
+  public ReadZeebeWorkerValue readZeebeWorkerValue(final ZeebeExpressionResolver resolver, final ZeebeClientProperties zeebeClientProperties) {
```
berndruecker, line 23: I don't like that we need the properties now available here - does not smell super good. However, I also don't have another great idea to get rid of it 🤔

README.md
berndruecker: Also I wonder if 
`zeebe.client.worker.workers`
is the most intuitive way of naming this - but out of good ideas at the moment

client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
```diff
@@ -586,4 +599,9 @@ public JsonMapper getJsonMapper() {
     return jsonMapper;
   }
 
+  @Override
+  public Map<String, WorkerConfiguration> getWorkersConfiguration() {
```
berndruecker, line 603: Why do we name it differently in the interface than in the config-object (=properties/yaml) file? I think it should be the same to avoid any confusion.

If we want to separate the usage from the definition, I would probably see 

`getWorkerConfiguration(String workertype)`

as a useful interface/abstraction - WDYT?

Then this would probably hint towards using `zeebe.client.worker.config['foo'].enabled=false` as a property name?
Or would `zeebe.client.worker.foo.enabled=false` work somehow? Not sure about Spring here - do you know? 

I wonder if we make any checks if people define multiple workers for one type 🤔

README.md
```diff
@@ -382,6 +382,28 @@ class MyConfiguration {
 }
 ```
 
+### Disable worker
+
+If you need to disable worker you can do it using properties:
```
berndruecker, line 387: It would be helpful to mention why this feature is helpful (I found the comment in the issue helpful to understand the why)

berndruecker (commented):
Thanks a lot for your diligent work! Looks generally good, but I had a couple of questions (see below)

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/value/factory/ReadAnnotationValueConfiguration.java
aivinog1: We can create a custom interface and extract the worker's configuration method to this new interface. And then implement this interface in the `ZeebeClientConfigurationProperties` class 🤔  I will try it and propose it in the PR :)

aivinog1: I've played a little with the code and maybe shall we introduce some sort of `ZeebeWorker` annotation customizer interface? So, it would work something like this:
1. We implement this interface in the `ZeebeClientConfigurationProperties` with default order.
2. We autowire a list of these interface's implementations into the `ZeebeWorkerAnnotationProcessor` class
3. We go through this list
Thus, end users get the ability to provide their own customizations if needed and in our code, there will be no dependency around properties at all. Also, end users could override the order (via `@Order` annotation) of these customizations and could control this process thoroughly.

berndruecker: I definitely like to have the switch if a worker is enabled or not in the `ZeebeWorkerAnnotationProcessor` - this feels like the right place. Having an Customizer interface sounds OK - maybe you could push the code you (@aivinog1) have in mind than I am happy to have a look?

aivinog1: Hey @berndruecke! I've changed the approach: right now I'm using customizers instead of direct dependency on the `ZeebeClientConfiguration`. Can you take a look? :) From my perspective, it is cleaner than the previous one :)

client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
aivinog1: Right now there is no direct dependency and we can use the map directly from the `Worker` class :)

README.md
aivinog1: Based on https://github.com/camunda-community-hub/spring-zeebe/pull/221#discussion_r965570205 I think that the  best name for it should be something like this: 
```properties
zeebe.client.worker.override.some-worker-type.enabled=false
```
WDYT?

README.md
aivinog1: Done 👍

aivinog1:
Hey @berndruecker! This one is waiting for your review :)

aivinog1:
Hey @berndruecker! I'm so sorry about bothering you, but can you take a look, please? 🙂

berndruecker (approved):

berndruecker:
Howdy @aivinog1 - thanks once again. I really like this design much more! I will merge this now and then probably slightly adjust the readme. I have one or two other things I wanted to add, and then I could also go ahead and do a release with the change

aivinog1:
Thanks @berndruecker! :)

PR #229, opened 2022-09-27
New annotation names and defaults

falko:
Based on experience of using Spring Zeebe, Bernd and I are proposing a refactoring of annotations names and their default values.

### DIFF
--- a/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/JobWorker.java
+++ b/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/JobWorker.java
@@ -9,7 +9,10 @@ import java.lang.annotation.Target;
 @Target(ElementType.METHOD)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-public @interface ZeebeWorker {
+public @interface JobWorker {
+  // Alternative naming ideas:
+  // JobSubscription
+  // JobListener
 
   String type() default ""; // set to empty string which leads to method name being used (if not ${zeebe.client.worker.default-type}" is configured) Implemented in ZeebeWorkerAnnotationProcessor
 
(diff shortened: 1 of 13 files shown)


falko:
@berndruecker the solution architects are happy with the current state including keeping the slightly longer but more explicit `@VariablesAsType`

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/JobWorker.java
```diff
@@ -9,7 +9,10 @@
 @Target(ElementType.METHOD)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-public @interface ZeebeWorker {
+public @interface JobWorker {
```
chDame, line 12: Rather than renaming, I would duplicate and mark the ZeebeWorker as deprecated  with a link to the new annotation explaining the changing behaviors.

chDame: Same remark for other annotations :)

falko:
This example should change too: https://github.com/camunda/camunda-platform-tutorials/blob/main/orchestrate-microservices/worker-java/src/main/java/io/camunda/getstarted/tutorial/Worker.java

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/JobWorker.java
berndruecker: +100! The idea with that branch was simply to show the target.
Really doing it involves some more work also to make sure deprecated anootations are still processes normally.

falko:
@berndruecker stakeholders are fine, too.
@menski FYI

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/JobWorker.java
chDame: wonderfull :)

berndruecker:
Moved to https://github.com/camunda-community-hub/spring-zeebe/issues/239

PR #236, opened 2022-10-07
fixes ZeebeVariable deserialization

bulivlad:
Fixes the deserialization of complex types using an ObjectMapper when annotating a worker job method with `@ZeebeVariable`

PR #237, opened 2022-10-10
fixes ZeebeVariable deserialization using JsonMapper

bulivlad:
Fixes the deserialization of complex types when annotating a worker job method with @ZeebeVariable by using the `JsonMapper` bean when exists in the context, falling back to a plain `ObjectMapper` otherwise

Supersede #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,7 +403,9 @@ If you need to customize the ObjectMapper that the Zeebe client uses to work wit
 class MyConfiguration {
   @Bean
   public JsonMapper jsonMapper() {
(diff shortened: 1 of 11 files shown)


bulivlad:
just for visibility - not really happy on how the `JsonMapper` field is handled in `ZeebeClientConfigurationProperties` but as long as `ZeebeClientConfiguration` in `zeebe-client-java` declares `JsonMapper getJsonMapper();` I cannot think of any better way to handle it. Any ideas on this one are more than welcome :)

berndruecker:
LGTM

berndruecker:
I agree that the handling of the JsonMapper is not optimal - but I would not touch it for the moment--

Issue #240, opened 2022-10-12
8.0.11 ZeebeClientStarterAutoConfiguration required a bean of type 'io.camunda.zeebe.client.api.JsonMapper' that could not be found.


zxiaozhou:
update to 8.0.11 not start
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 0 of method builder in io.camunda.zeebe.spring.client.config.ZeebeClientStarterAutoConfiguration required a bean of type 'io.camunda.zeebe.client.api.JsonMapper' that could not be found.


Action:

Consider defining a bean of type 'io.camunda.zeebe.client.api.JsonMapper' in your configuration.

```

berndruecker:
Thanks for reporting @zxiaozhou - actually this bean should be optional, let me double check why this is not the case and come back. You could keep running 8.0.9 for the time being

berndruecker:
This is pretty wired. The bean is missing here: https://github.com/camunda-community-hub/spring-zeebe/blob/8.0.11/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/config/ZeebeClientStarterAutoConfiguration.java#L35
But actually created here: https://github.com/camunda-community-hub/spring-zeebe/blob/8.0.11/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/config/ZeebeClientStarterAutoConfiguration.java#L70

@bulivlad: Do you have any idea why this is failing? 

@zxiaozhou: I can't reproduce the problem with some of my local Spring Boot apps, they all run fine with 8.0.11 - any chance you share some part of your code that I could see how the application is wired together?

berndruecker:
I agree - really wired. Waiting for @zxiaozhou now - and decided that this will not block the 8.1.0 release.

zxiaozhou:
@berndruecker I get an error when I add the spring-cloud-starter-stream-rabbit dependency
the  demo
[spring-zeebe-demo.zip](https://github.com/camunda-community-hub/spring-zeebe/files/9782199/spring-zeebe-demo.zip)

zxiaozhou:
@berndruecker I get an error when I add the spring-cloud-starter-stream-rabbit dependency
the  demo
[spring-zeebe-demo.zip](https://github.com/camunda-community-hub/spring-zeebe/files/9782199/spring-zeebe-demo.zip)

zxiaozhou:
log info
```
[...]

PR #242, opened 2022-10-14
fixed zeebe jsonMapper bean name being duplicated with cloud stream jsonMapper name #240

zxiaozhou:
Fixed startup failure caused by the same zeebe jsonMapper bean name and cloud stream jsonMapper name

Supersede https://github.com/camunda-community-hub/spring-zeebe/issues/240


berndruecker:
@bulivlad WDYT?

bulivlad:
thanks @zxiaozhou for the changes. finally got some time to look into this one. the fix looks fine to me as long as we are fine to restrict others to override the `JsonMapper` bean using `zeebeObjectMapper` name (which I'm not a fan of) - otherwise the overriding bean will not be injected. the tests also needs some changes to pass.

Also proposed a pull request here #243 that should fix the issue and let others to pick whatever bean name they want - can be closed if we are fine with this PR. 

@berndruecker your call here

berndruecker:
I merged #243 into master. @zxiaozhou - could you double check if that fixes your problem properly? 


PS: I can't merge PR's without signed SLA...

zxiaozhou:
@berndruecker It has been solved

berndruecker:
Great to hear! Thx again for reporting it so quickly

PR #253, opened 2022-11-09
Connector handler

tmetzke:
Reuse the Connector SDK's `ConnectorJobHandler` for Connector function execution.


berndruecker:
Howdy @tmetzke! Thanks! I am generally fine with this for now. I would still like to harmonize exception/retry/error handling for JobWorkers and Connectors - so remove that separation again. This would allow Java developers to write Connector Functions and use them like workers  - with the potential of easier reuse. But we can postpone this for now and release 0.3.0 first

PR #260, opened 2022-11-11
chore(connector-runtime): created shaded artifact

tmetzke:


berndruecker:
Ok with me if you prefer it to have this way

berndruecker (approved):

tmetzke:
It is what is documented everywhere and it's more obvious I think 🙂

Issue #268, opened 2022-11-23
Add metrics to Connector Runtime formerly written by the Cloud Bridge


berndruecker:
Splitted this into two parts:

- [x] Replace https://github.com/camunda/cloud-connector-bridge/blob/main/src/main/java/io/camunda/connectorbridge/invoker/ConnectorInvokerMetrics.java
- [x] Replace https://github.com/camunda/cloud-connector-bridge/blob/main/src/main/java/io/camunda/connectorbridge/zeebe/JobMetrics.java

Metrics are available via Spring Actuator on http://localhost:8080/actuator/metrics/:
* `camunda.connector.outbound.invocations`
* `camunda.connector.inbound.invocations`
* `camunda.job.invocations`

Work in: https://github.com/camunda-community-hub/spring-zeebe/pull/269

berndruecker:
The Bridge metric layout [reuses the same counter for all different type of things (executed, failed, ...)](https://github.com/camunda/cloud-connector-bridge/blob/main/src/main/java/io/camunda/connectorbridge/invoker/ConnectorInvokerMetrics.java#L14-L19). Even if different tags are given, those just lead to a list of possible tags, see e.g. 

![grafik](https://user-images.githubusercontent.com/1217596/203730091-f6412754-5721-4f0c-9066-76155dddbfaa.png)

This is inline to what e.g. HTTP does as metrics:

![grafik](https://user-images.githubusercontent.com/1217596/203731002-fd05fcf6-61b9-49c0-9ac2-7142d4992161.png)

As the bridge only handled one connector, you got a number for its connector type from the bridge. The new runtime will handle multiple connectors, thus a counter per connector type makes more sense to me. I will add this to the name of the counter.

So
* camunda.connector.invocations.AWSLambda
* camunda.connector.invocations.HTTPJSON
* ...

![grafik](https://user-images.githubusercontent.com/1217596/203735337-f348ef1e-f61d-4c47-a418-b6b0546ff9a3.png)

tmetzke:
Synced offline, results:
* We'll keep name to metric itself, not involve Connector name in there
* We'll use tags/labels as before in the Bridge
* We'll include job metrics in Spring Zeebe as well
* We'll create a Micrometer-based implementation in Spring Zeebe Starter, conditionally activated when Actuator is available
* We'll include Actuator in the Spring Zeebe Connector Runtime application

berndruecker:
[...]

PR #269, opened 2022-11-23
Add connector metrics

berndruecker:
Will fix https://github.com/camunda-community-hub/spring-zeebe/issues/268


berndruecker:
@tmetzke Now We are exposing metrics for every outbound connector.

I might also have a quick look into inbound - I Wondered if this should influence metric names (eg changing to `camunda.connector.outbound.HTTPJSON.invocations`

so that we could also have the same connector on inbound, maybe also counting different things, e.g.
`camunda.connector.inbound.FTP.polls`
`camunda.connector.inbound.FTP.invocations`
...

client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/actuator/MicrometerMetricsRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+public class MicrometerMetricsRecorder implements MetricsRecorder {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private final MeterRegistry meterRegistry;
+
+  private final Map<String, Counter> counterActivated = new HashMap<>();
+  private final Map<String, Counter> counterCompleted = new HashMap<>();
+  private final Map<String, Counter> counterFailed = new HashMap<>();
+  private final Map<String, Counter> counterBpmnError = new HashMap<>();
```
tmetzke, line 27: Isn't it enough to just have Counter objects here for each type? Either that or one map with a counter per action.

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/connector/ConnectorConfiguration.java
```diff
@@ -15,9 +13,23 @@
...
 
+  public static class OnMissingMetricsRecorder implements Condition {
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+      return context.getBeanFactory().getBeanNamesForType(MetricsRecorder.class).length<=0;
+    }
+  }
+
+  @Bean
+  @Conditional(value=OnMissingMetricsRecorder.class)
```
tmetzke, line 24: Can we also simply do a [ConditionalOnMissingBean](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnMissingBean.html) here? As soon as someone creates a custom bean of the `MetricsRecorder` class, this here should be ignored. I think. 🙂

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/SpringConnectorJobHandler.java
```diff
@@ -3,30 +3,44 @@
...
+    super(connectorConfiguration.getFunction(), secretProvider);
+    this.connectorConfiguration = connectorConfiguration;
+    this.metrics = metrics;
     this.secretProvider = secretProvider;
     this.commandExceptionHandlingStrategy = commandExceptionHandlingStrategy;
   }
 
+  @Override
+  public void handle(JobClient client, ActivatedJob job) {
+    metrics.increaseActivated(MetricsRecorder.METRIC_NAME_OUTBOUND_CONNECTOR, connectorConfiguration.getType());
```
tmetzke, line 30:
```suggestion
    metrics.increaseActivated(MetricsRecorder.METRIC_NAME_JOB, connectorConfiguration.getType());
```

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/SpringConnectorJobHandler.java
```diff
@@ -3,30 +3,44 @@
...
+    super.handle(client, job);
+  }
+
   @Override
   protected void failJob(JobClient client, ActivatedJob job, Exception exception) {
+    metrics.increaseFailed(MetricsRecorder.METRIC_NAME_OUTBOUND_CONNECTOR, connectorConfiguration.getType());
```
tmetzke, line 36:
```suggestion
    metrics.increaseFailed(MetricsRecorder.METRIC_NAME_JOB, connectorConfiguration.getType());
```

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/SpringConnectorJobHandler.java
```diff
@@ -35,6 +49,7 @@ protected void throwBpmnError(JobClient client, ActivatedJob job, BpmnError valu
   }
 
   @Override protected void completeJob(JobClient client, ActivatedJob job, ConnectorResult result) {
+    metrics.increaseCompleted(MetricsRecorder.METRIC_NAME_OUTBOUND_CONNECTOR, connectorConfiguration.getType());
```
tmetzke, line 52:
```suggestion
    metrics.increaseCompleted(MetricsRecorder.METRIC_NAME_JOB, connectorConfiguration.getType());
```

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/connector/ConnectorConfiguration.java
berndruecker: We can't - as the current spring-zeebe module does not depend on Spring Boot - only Spring Context, hence the workaround. I think it should work for all relevant cases - but always happy to improve later on if we learn that it doesn't. We might also double check if somebody needs a pure spring (non spring boot) dependency - but I find it hard to find that out

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/SpringConnectorJobHandler.java
berndruecker: The original code is correct /intended. The ConnectorJobHandler is only called for connectors - so it should count connectors - I just wondered if we should maybe use the connector name instead of the type (but then we would miss the version).

client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/actuator/MicrometerMetricsRecorder.java
berndruecker: Yeah  -the latter. The recorder can handle multiple metric names, currently one for JOB and one cor CONNECTOR_INVOCATIONS. I am pretty sure we will also add something for INBOUND. I agree though, that it might not be a design super easy to follow - let me adjust it

PR #271, opened 2022-11-25
Made activationCondition and variableMapping optional for webhook

berndruecker:
Those two configurations are not required by the webhook (the code can work with null values and will apply defaults). So they should not be enforced when parsing the config. 

OK @igpetrov?


igpetrov (approved):
LGTM

PR #276, opened 2022-11-28
Adds JavaDocs to @JobWorker

namero999:
Added docs to `@JobWorker` as this is one of the primary interfaces to create workers and they lack important info such as unit of measurement for timeouts.

Most of the wording has been inherited from `io.camunda.zeebe.client.api.worker.JobWorkerBuilderStep1`.

Please double check the UoM for `requestTimeout` (the only one in seconds as opposed to milliseconds). This has been inferred by looking at `JobWorkerManager.java#L77`.


berndruecker:
Thanks @namero999 !

Issue #293, opened 2022-12-12
[Bug] Failed to serialize OffsetDateTime


Oleksiivanov:
**Describe the bug**

When zeebe tries to serialize an object with [OffsetDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html) ObjectMapper throw exception : 
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.OffsetDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: java.util.HashMap["result"]->com.microsoft.graph.models.ChatMessage["createdDateTime"])
```

**Steps to reproduce**

1. Create a new BPMN diagram with Microsoft Teams connector template.
2. Choose Conversation type `chat` , and Method `Send message to chat` in section `operation`
3. Fill required fields
4. Deploy diagram and start instance 
5. in `operate` we can see error : 

<details><summary>Click for see Full Stacktrace</summary>
 <p>

```
java.lang.RuntimeException: io.camunda.zeebe.client.api.command.InternalClientException: Failed to serialize object '{result=com.microsoft.graph.models.ChatMessage@5063940a}' to json
	at io.camunda.zeebe.spring.client.jobhandling.SpringConnectorJobHandler.failJob(SpringConnectorJobHandler.java:37)
	at io.camunda.connector.runtime.util.outbound.ConnectorJobHandler.handle(ConnectorJobHandler.java:94)
	at io.camunda.zeebe.spring.client.jobhandling.SpringConnectorJobHandler.handle(SpringConnectorJobHandler.java:30)
	at io.camunda.zeebe.spring.client.jobhandling.JobHandlerInvokingSpringBeans.handle(JobHandlerInvokingSpringBeans.java:76)
	at io.camunda.zeebe.client.impl.worker.JobRunnableFactory.executeJob(JobRunnableFactory.java:44)
	at io.camunda.zeebe.client.impl.worker.JobRunnableFactory.lambda$create$0(JobRunnableFactory.java:39)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
[...]

PR #294, opened 2022-12-12
fix(JsonMapper): register Module JavaTimeModule for parse OffsetDateTime

Oleksiivanov:
register Module JavaTimeModule in zeebeObjectMapper for parse OffsetDateTime.

close:
https://github.com/camunda-community-hub/spring-zeebe/issues/293

### DIFF
--- a/connector-runtime/pom.xml
+++ b/connector-runtime/pom.xml
@@ -20,6 +20,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+    </dependency>
+    <dependency>
       <groupId>io.camunda.connector</groupId>
       <artifactId>connector-runtime-util</artifactId>
     </dependency>
--- a/client/spring-zeebe-starter/pom.xml
+++ b/client/spring-zeebe-starter/pom.xml
@@ -30,10 +30,6 @@
       <artifactId>spring-boot-starter-actuator</artifactId>
       <optional>true</optional>
     </dependency>
-     <dependency>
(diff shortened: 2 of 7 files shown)


connector-runtime/pom.xml
```diff
@@ -41,6 +41,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
berndruecker, line 44: Shouldn't this come as transitive dependency from the starter (and thus not be necessary?)

berndruecker (commented):
I can have a look or merge it later today (afternoon). Is that sufficient @Oleksiivanov ?

ivanjovic:
I stumbled across the same problem today, maybe it could be an idea to use the existing Objectmapper from the Spring Context (as it already comes with the missing jackson module).
It would make providing a custom Objectmapper easier (just override the default `Objectmapper` Spring Bean with your custom one and the `JsonMapper` would pick it up automatically). This would look something like the following (this worked like a charm in my demo app):
```java
    @Bean
    public JsonMapper jsonMapper(final ObjectMapper objectMapper) {
        return new ZeebeObjectMapper(objectMapper);
    }
```

Don't know if I missed something but that should make the config more flexible.

berndruecker:
I think @ivanjovic's proposal is a good hint for how Spring Boot is typically handling those situations. So I adjust the code accordingly and will also move the `jackson-datatype-jsr310` dependency to the connector-runtime only - as I would love to leave it to spring-zeebe users to decide what jackson dependencies they actually want to have. Testing this now.

berndruecker:
Thanks again @ivanjovic - your comment was spot on and how it should be done using Spring Boot. I adjusted the PR accordingly and will merge it now.

@Oleksiivanov Maybe you can retest the teams connector once you are back?

PR #297, opened 2022-12-15
fix(metrics): Restore default actuator metrics (e.g. `jvm`)

### LINKED ISSUE #296: Default metrics provided by `spring-boot-actuator` don't work
### Issue description

`MetricsRecorder` bean declaration in `spring-zeebe-starter` breaks the default metrics auto configurations provided by Spring Boot Actuator.
https://github.com/camunda-community-hub/spring-zeebe/blob/47e151387040920b30c8914297fe084bc37cb00c/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/actuator/ZeebeActuatorConfiguration.java#L23

Apparently, the `MeterRegistry` bean is eagerly injected, which breaks the initialization of other metrics. As a result, for, example, the `jvm` metrics are missing in the actuator endpoint:
Screenshot 2022-12-15 at 13 46 25

### Possible solution

* Inject `MeterRegistry` bean with `@Lazy`

Similar issues with example solutions:
* https://github.com/spring-projects/spring-boot/issues/26630
* https://stackoverflow.com/questions/52148543/spring-boot-2-actuator-doesnt-publish-jvm-metric/53119998#53119998

chillleader:
related to #296


tmetzke (approved):
Looks good to me as a quick fix, still feels like a workaround to me though 😀 
Do we have any proof that this works? Can you share screenshots or anything?

berndruecker:
Awesome - thx @chillleader

PR #300, opened 2022-12-20
fix(metrics): include jobType tag in counter key

### LINKED ISSUE #299: Incorrect behavior of `camunda_connector_outbound_invocations_total` metric
Metric `camunda_connector_outbound_invocations_total` always has the same `type` tag.

## How to reproduce

1. Model a process that uses at least 2 different connectors and execute it.
2. Call `/actuator/prometheus` endpoint and check the `camunda_connector_outbound_invocations_total` metric.

## Expected behavior

There should be one entry of this metric per connector, with the corresponding `type` tag.

## Actual behavior

The screenshot below the state of application metrics after executing a process with 2 different connectors (Slack & Google Drive). Since the slack connector was executed first, tag `io.camunda:slack:1` is applied to all job invocations no matter the actual connector type.

Screenshot 2022-12-20 at 12 01 32

chillleader:
Provides fix of the `camunda_connector_outbound_invocations_total` metric (and subsequently of the similar metric for inbound). See related issue for context.

closes #299


markfarkas-camunda (approved):
Nice! Works like a charm! 👍

PR #304, opened 2022-12-27
feat(metrics): collect execution time metrics

markfarkas-camunda:
https://github.com/camunda/team-connectors/issues/289 

related to #305

### DIFF
--- a/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerInvokingSpringBeans.java
+++ b/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerInvokingSpringBeans.java
@@ -73,7 +73,10 @@ public class JobHandlerInvokingSpringBeans implements JobHandler {
   public void handle(JobClient jobClient, ActivatedJob job) throws Exception {
     if (outboundConnectorConfiguration!=null) {
       LOG.trace("Handle {} and execute connector {}", job, outboundConnectorConfiguration);
-      new SpringConnectorJobHandler(outboundConnectorConfiguration, secretProvider, commandExceptionHandlingStrategy, metricsRecorder).handle(jobClient, job);
+      metricsRecorder.executeWithTimer(job.getType(), () -> {
+        new SpringConnectorJobHandler(outboundConnectorConfiguration, secretProvider, commandExceptionHandlingStrategy, metricsRecorder).handle(jobClient, job);
+      });
+
     } else { // "normal" @JobWorker
       // TODO: Figuring out parameters and assignments could probably also done only once in the beginning to save some computing time on each invocation
       List<Object> args = createParameters(jobClient, job, workerValue.getMethodInfo().getParameters());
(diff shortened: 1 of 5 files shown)


client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/metrics/DefaultNoopMetricsRecorder.java
```diff
@@ -12,4 +12,9 @@ public class DefaultNoopMetricsRecorder implements MetricsRecorder {
   public void increase(String metricName, String action, String type) {
     // ignore
   }
+
+  @Override
+  public void executeWithTimer(String metricName, Runnable methodToExecute) {
+    // ignore
```
chillleader, line 18: Should we add `methodToExecute.run()` here? Otherwise it seems like the method won't be executed at all.

markfarkas-camunda: Very well spotted, thank you 👍 I fixed it.

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerInvokingSpringBeans.java
```diff
@@ -73,7 +73,10 @@ public JobHandlerInvokingSpringBeans(ZeebeWorkerValue workerValue,
   public void handle(JobClient jobClient, ActivatedJob job) throws Exception {
     if (outboundConnectorConfiguration!=null) {
       LOG.trace("Handle {} and execute connector {}", job, outboundConnectorConfiguration);
-      new SpringConnectorJobHandler(outboundConnectorConfiguration, secretProvider, commandExceptionHandlingStrategy, metricsRecorder).handle(jobClient, job);
+      metricsRecorder.executeWithTimer(job.getType(), () -> {
```
berndruecker, line 76: Two questions:
1. Why do we limit the time recording to connectors - I think this would be valuable to any Job Worker - or WDYT?
2. If we limit it to connectors, why not do it within the SpringConnectorJobHander as this already implements the counter - and thus we would rip apart metrics reporting in two places.

That's said I think we can also merge it right now and probably clean up later - as my guess is that we need some cleanup in this area sooner or later anyway (and it is at least relatively clear to understand)

berndruecker (approved):
I added a comment / question but I would also be OK with just merge it now to make progess. WDYT?

client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerInvokingSpringBeans.java
markfarkas-camunda: Valid points. Our initial though was to gather metric information only about the connectors, and that might be easier to achieve with Grafana if we do not have all Job Worker's execution time. I think we should merge it know and we can clean this up later on. My suggestion would be to have the execution time for all Job Worker in the future. I created a follow up ticket: https://github.com/camunda/team-connectors/issues/304

berndruecker: Make sense to me! For connectors having it for all Job workers would not change anything - as the connector runtime does not run any other job worker :-)

Issue #305, opened 2022-12-27
Add new metrics: thread pool, job execution time


chillleader:
### Metrics to consider:
* Job execution time
* Zeebe Client thread pool state

Since all workers share a common thread pool, it's interesting to know the state of this thread pool to adjust its size accordingly.

berndruecker:
@chillleader I just merged https://github.com/camunda-community-hub/spring-zeebe/pull/363 where the new way of configuration is added, so that you can inject the right things. Does this help to implement it (or is it already done 🤔 )

chillleader:
Nice, thank you Bernd

berndruecker:
Is this already done or should it be kept open?

chillleader:
Looks like everything is done here 👍

Issue #275, opened 2022-11-28
Switch from `@EnableZeebeClient` to Spring Boot Auto-Configuration


berndruecker:
I recently [added Metrics to Spring-Zeebe](https://github.com/camunda-community-hub/spring-zeebe/issues/268) and wanted that the user application can bring its own implementation of a MetricRecorder. Therefore, the core library would ship with a NoopMetricsRecorder. I tried to use Spring Configuration magic for this, to only create the NoopMetricsRecorder if no other MetricsRecorder is defined (e.g. @ConditionOnMissingBean). This did not work as expected and it took me a while to figure out the root cause, which is very well explained here: https://github.com/spring-projects/spring-boot/issues/18228

Bottom-line:
* The design using an `@EnableZeebeClient` annotation bypasses the Spring Auto Configuration mechanism
* This means, you cannot easily swap out beans by the mechanisms documented by Spring (or found via Google) - at the same time it is not very obvious why those solutions do not have any effect
* For my MetricsRecorder at hand I now use the workaround `@Autowired(required = false) MetricsRecorder` and create the Noop only if null is handed in. This works, but does not feel right.

Now I think we should change spring-zeebe:
* Deprecate `@EnableZeebeClient`
* Add `spring.factories` in meta-inf to automatically enable the Zeebe Client if the library is on the classpath
* Add a configuration option `zeebe.client.enabled` which can be set to `false` to disable the client if needed

Why?
* I don't see any situation where you add spring-zeebe to the classpath but not want the client to be enabled
* Using Spring Auto Configuration would ease up a lot of configuration work and make it easier understandable. We already hit problems around this in the past.

Does anybody see any problems with this?
8.2.x would be a good candidate to make such a change. It would generally work backwards compatible, unless somebody has an application that pulls in spring-zeebe-starter, but does not add @EnableZeebeClient. We could log a BIG WARNING in such a case to raise awareness for those situations (which should be very rare).

One open question is:
* Can we make spring-zeebe dependent on Spring Boot? This might be necessary to use auto configuration. The result is, that spring-zeebe would always depend on Spring Boot, right now spring-zeebe depends on spring-context and only spring-zeebe-starter depends on spring-boot.

berndruecker:
Any thoughts (maybe @lzgabel @aivinog1 @zambrovski @jangalinski @falko)? 
* Can we remove `@EnableZeebeClient` and auto start the client always (unless `zeebe.client.enabled=false`)?
[...]

PR #306, opened 2022-12-27
Rework Zeebe client configuration & add thread pool metrics

chillleader:
### 1. Create Zeebe Client using `ZeebeClientConfiguration` (instead of `ZeebeClientBuilder`)

To introduce thread pool metrics, first and foremost we need to have access to the `ScheduledExecutorService` used by Zeebe Client.
This is impossible if we only use `ZeebeClientBuilder`, since it doesn't allow us to configure the executor service.
`ZeebeClientImpl` class, on the other hand, allows us to create an instance with a custom executor service.

I also noticed that `ZeebeClientConfigurationProperties` implements `ZeebeClientConfiguration`, which can be directly used to create a client object. This also allows us to avoid re-wrapping all our properties in a Builder.

**Concern** - backward compatibility is an issue in this specific scenario (though it doesn't seem likely):
* Someone has `spring-zeebe` in the classpath, but doesn't use the spring boot starter, and also relies on the existing mechanism by providing a custom `ZeebeClientBuilder` bean in context. 

### 2. Thread pool metrics

* On creation of `ScheduledExecutorService`, `ExecutorServiceMetrics` instance is created
* New metrics become available:
  * executor_completed_tasks_total
  * executor_pool_size_threads
  * executor_active_threads
  * executor_queued_tasks
 
Screenshot 2022-12-27 at 17 27 04

related to #305, camunda/team-connectors#288

**P.S.** This still needs some polishing and new tests, I will provide those if we agree on the chosen approach.

### DIFF
--- a/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/config/ZeebeClientStarterAutoConfiguration.java
+++ b/client/spring-zeebe-starter/src/main/java/io/camunda/zeebe/spring/client/config/ZeebeClientStarterAutoConfiguration.java
@@ -3,22 +3,28 @@ package io.camunda.zeebe.spring.client.config;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.zeebe.client.ZeebeClientBuilder;
 import io.camunda.zeebe.client.api.JsonMapper;
-import io.camunda.zeebe.client.impl.ZeebeClientBuilderImpl;
 import io.camunda.zeebe.client.impl.ZeebeObjectMapper;
 import io.camunda.zeebe.spring.client.actuator.ZeebeActuatorConfiguration;
 import io.camunda.zeebe.spring.client.annotation.customizer.ZeebeWorkerValueCustomizer;
+import io.camunda.zeebe.spring.client.lifecycle.ZeebeClientObjectFactory;
+import io.camunda.zeebe.spring.client.lifecycle.ZeebeClientObjectFactoryImpl;
 import io.camunda.zeebe.spring.client.properties.PropertyBasedZeebeWorkerValueCustomizer;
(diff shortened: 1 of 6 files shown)


berndruecker:
Howdy @chillleader: regarding changes & backwards compatibility - sounds like a great thing to do with 8.2.0 (as we will have changes related to https://github.com/camunda-community-hub/spring-zeebe/issues/275 anyway). Could we postpone it till then or do you see this more urgent? I think then I also like that it simplifies things a bit (less re-wrapping required).

chillleader:
Yeah, let's hold it till 8.2.0 👍

berndruecker:
@chillleader - just FYI: I will look at this today or tomorrow and either adjust this PR or probably also create a new one based on latest `main`

chillleader:
Thank you! ⭐

berndruecker:
I found one possible difference: https://github.com/camunda/zeebe/blob/main/clients/java/src/main/java/io/camunda/zeebe/client/impl/ZeebeClientBuilderImpl.java#L311 applies overriding certain Zeebe properties with environment variables defined in the Java ZeebeClient.

Now we have three options:
- Provide a PR to the ZeebeClient to allow swapping the `ScheduledExecutorService` but keep the logic as is
- Rewrite the `applyOverrides()` logic in Spring Zeebe
- Ignore those environment variable names in Spring Zeebe and don't support those names

Any opinion @chillleader? 

I like (3) most, but don't want to cause any trouble so I think we would go for (2).

chillleader:
I like (1) the most, as it seems like a missing feature in Zeebe Client. But (2) also seems fine, as clearly it may be easier to just make the change here 🙂

berndruecker:
I went for (2) as this could be a rabbit hole on how to configure/setup the ZeebeClient in Java and also that was a good hook to apply certain legacy logic anyway (https://github.com/camunda-community-hub/spring-zeebe/issues/350)

PR #311, opened 2023-01-10
added necessary classes for inbound connector/subscription

berndruecker:
Quick first stab at a possible inbound connector support (subscription) solution for discussion.

Requires https://github.com/camunda/connector-sdk/pull/334 to be built properly


berndruecker:
Will be done with https://github.com/camunda-community-hub/spring-zeebe/pull/317

PR #317, opened 2023-01-17
Inbound Intermediate Events (Spring Zeebe changes)

chillleader:
⚠️ Build depends on https://github.com/camunda/connector-sdk/pull/346

## Description

Changes required for inbound intermediate events.

### Change list
- Implementation of message correlation under `ProcessCorrelationPoint` abstraction (see SDK PR)
- Rearrangement of process instance creation code to match the same abstraction

## Related issues

https://github.com/camunda/team-connectors/issues/308


connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/event/MessageInboundTarget.java
```diff
@@ -0,0 +1,99 @@
...
+import io.camunda.zeebe.client.ZeebeClient;
+import io.camunda.zeebe.client.api.response.PublishMessageResponse;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MessageInboundTarget extends InboundConnectorTarget {
+  private static final Logger LOG = LoggerFactory.getLogger(MessageInboundTarget.class);
+
+  private final String messageName;
```
berndruecker, line 16: I wonder if it make sense to have the message properties here and the startEvent properties in the super class - should be both in subclasses I guess?

That's said maybe we should have `AbstractMEssageInboundTarget` with those properties in the SDK and only the "triggerEvent" part here? Same for StartEvent

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/event/MessageInboundTarget.java
```diff
@@ -0,0 +1,99 @@
...
+      return -1;
+    }
+    MessageInboundTarget other = (MessageInboundTarget) o;
+    if (!correlationKey.equals(other.correlationKey)) {
+      return correlationKey.compareTo(other.correlationKey);
+    }
+    return messageName.compareTo(other.messageName);
+  }
+
+  @JsonTypeName("MESSAGE")
```
berndruecker, line 65: Would Jackson dependency here be probably sufficient?

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/event/StartEventInboundTarget.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+    StartEventInboundTarget other = (StartEventInboundTarget) o;
+    if (!bpmnProcessId.equals(other.bpmnProcessId)) {
+      return bpmnProcessId.compareTo(other.bpmnProcessId);
+    }
+    return Integer.compare(version, other.version);
+  }
+
+  @JsonTypeName("START_EVENT")
+  public static class Response extends InboundConnectorResult {
```
berndruecker, line 66: personally not a big fan of static inner classes - just saying :-) Not sure if we have general preferences...

berndruecker (commented):
I generally like the direction very much! I made some minor comments and have not looked through every single line of code - but I think it is generally quite close to mergable :-)

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/event/StartEventInboundTarget.java
chillleader: I don't have a strong preference for either of the options as long as it looks readable, so no problem :)

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/event/MessageInboundTarget.java
chillleader: I reorganized the whole class hierarchy here & in the SDK, I hope it feels a bi more robust now 🙂

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/registry/InboundConnectorRegistry.java
```diff
@@ -69,7 +75,9 @@ public void registerWebhookConnector(InboundConnectorProperties properties) {
     WebhookConnectorProperties webhookConnectorProperties =
         new WebhookConnectorProperties(properties);
 
-    String bpmnId = webhookConnectorProperties.getBpmnProcessId();
+    ProcessCorrelationPoint correlationPoint = properties.getCorrelationPoint();
+    String bpmnId = correlationPoint.getBpmnProcessId();
```
berndruecker, line 79: OK - I think I get the idea why you want to have the processId in the CorrelationPoint. 

I still find this a little bit confusing. As an alternative we could add the relevant properties also to the InboundConnectorProperties directly? Then we would separate the two pieces of information: Process definition defining the inbound connector - and the correlation point itself (completely opaque here). 

But also just my gut feeling - feel free to ignore :-)

berndruecker (approved):
LGTM - just have one remark about "misusing" some properties on the correlation point

PR #318, opened 2023-01-20
Add zeebe client for test early in context

### LINKED ISSUE #313: Problem with the `ZeebeClient` at the Spring Context initialization phase
I'm creating a Spring Boot starter for the coworker project and I've decided that it should depend on Spring Zeebe.
I've such configuration class:

```kotlin
@Configuration
@AutoConfigureAfter(ZeebeClientStarterAutoConfiguration::class, ZeebeClientSpringConfiguration::class)
@ConditionalOnClass(ZeebeClientStarterAutoConfiguration::class, ZeebeClient::class, ZeebeClientSpringConfiguration::class)
open class CoworkerAutoConfiguration {

    @Bean
    open fun coZeebe(zeebeClient: ZeebeClient): Cozeebe {
        return zeebeClient.toCozeebe()
    }
}
```

The `ZeebeClient#toCozeebe` method looks like this:

```kotlin
fun ZeebeClient.toCozeebe() = CozeebeImpl(this. configuration)
```

So, I have to have ZeebeClient with configuration on the Spring Context Initialization phase. It means that the `ZeebeClientProxy` must have a ZeebeClient instance in the context initialization phase.

I think that my case isn't an exception, this applies to all users that doing something with `ZeebeClient` in the context initialization phase (deploying resources, starting instances, lookup for topology, etc.).

aivinog1:
I've decided that `ZeebeClient` should be firstly initialized in a Spring context. So, it requires that `ZeebeEngine` should be initialized before it as well.

Closes #313

### DIFF
--- a/test/common/src/main/java/io/camunda/zeebe/spring/test/AbstractZeebeTestExecutionListener.java
+++ b/test/common/src/main/java/io/camunda/zeebe/spring/test/AbstractZeebeTestExecutionListener.java
@@ -14,6 +14,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.test.context.TestContext;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Optional;
 
 /**
  * Base class for the two different ZeebeTestExecutionListener classes provided for in-memory vs Testcontainer tests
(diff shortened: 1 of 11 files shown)


aivinog1:
Hey @berndruecker 👋

Can you please take a look at this PR? I would like to see this in the future version of Spring Zeebe 🙂

Thank you 🙇

aivinog1:
Hello @berndruecker 👋

I'm sorry for bothering you, but can you take a look at this, please? 🙂

When you release a new version with this fix it unblocks my PR for the first version of Coworker. 

Thank you 🙇

test/common/src/main/java/io/camunda/zeebe/spring/test/AbstractZeebeTestExecutionListener.java
```diff
@@ -22,33 +25,28 @@ public class AbstractZeebeTestExecutionListener {
...
   private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  private ZeebeClient zeebeClient;
+  public <T extends ZeebeTestEngine> void setupWithZeebeEngine(TestContext testContext, T zeebeEngine, final UnaryOperator<T> startCallback) {
 
-  /**
-   * Registers the ZeebeEngine for test case in relevant places and creates the ZeebeClient
-   */
-  public void setupWithZeebeEngine(TestContext testContext, ZeebeTestEngine zeebeEngine) {
+    if (Optional.ofNullable(testContext.getAttribute("NOT_STARTED")).map(o -> (Boolean) o).orElse(false)) {
```
berndruecker, line 30: Small thing, but confuses me: I would name the attribute "STARTED" to avoid a double-boolean logic :-)

That's said I wonder why when called the first time, this is initialized with "NOT_STARTED" = False = STARTED? That means the engine is not set on the proxy the first time. I double checked and I think that is because you now initialize the proxy immediately when creating. This was cautiously changed - right? 

IF I understand it correctly, the first engine is now created on bean creation time - and then in the first test method nothing happens. In the following test methods, the engine is swapped. 

Engine creation is then not done in the before/after methods directly, but by the callback handed in - so you can control here when it is created. Is that a correct description of your idea?

berndruecker: @aivinog1

aivinog1: Hey @berndruecker! 
> That means the engine is not set on the proxy the first time. I double checked and I think that is because you now initialize the proxy immediately when creating. This was cautiously changed - right?
> IF I understand it correctly, the first engine is now created on bean creation time - and then in the first test method nothing happens. In the following test methods, the engine is swapped.
> Engine creation is then not done in the before/after methods directly, but by the callback handed in - so you can control here when it is created. Is that a correct description of your idea?
Yes, that is true, I make it intentionally and this is true about my idea :)

berndruecker:
With https://github.com/camunda-community-hub/spring-zeebe/issues/275 on the way I would close this PR - let me know if you object @aivinog1

Issue #324, opened 2023-02-01
Option to Dynamically Register JobWorker without annotations?


SaiprasadKrishnamurthy:
Hi , looking at the spring-zeebe project README and examples and I couldn’t find anything to dynamically create the JobWorker without the compile time annotations.

I’m trying to create a common library for Camunda/Zeebe where my services know nothing about the Workflow engine. It is pretty much agnostic and this library should take care of dynamically registering workers (specified by individual services) and the individual services just get a plain java call back with necessary payload and rest of the boiler plate is all handled by this common library.

I see this is only possible if there’s an alternative way to create job worker other than via @JobWorker annotation. Thoughts?

Kind of similar to Spring providing Dynamic registration mechanism to create Kafka Listener on the fly than @KafkaListener annotation.

Curious-Eye:
I think you can autowire ZeebeClientLifecycle and use it's method "newWorker()" to dynamically register workers.

berndruecker:
Hy @SaiprasadKrishnamurthy: You can register job workers via the JobWorkerManager, see how this is done internally after annotations are parsed: https://github.com/camunda-community-hub/spring-zeebe/blob/main/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/processor/ZeebeWorkerAnnotationProcessor.java#L115

This should solve your problem - or not?

PR #326, opened 2023-02-02
DynamicJobWorkerRegistrar to dynamically or programmatically register Job Workers

SaiprasadKrishnamurthy:
An alternative to @JobWorker annotation for a more programmatic worker registration:

```
DynamicJobWorkerRegistrar.newInstance(applicationContext)
                    .withJobName("boo")
                    .withJobType("boo")
                    .withHandler(taskReceivedEvent -> {
                        System.out.println(" ==> Received: " + taskReceivedEvent.getVariables());
                    }).build();
```

Or

```
DynamicJobWorkerRegistrar.newInstance(applicationContext)
                    .withJobName("boo")
                    .withJobType("boo")
                    .emitSpringApplicationEvent(true)
.build();

// In this case, this will emit TaskReceivedEvent as a Spring application Event and can be consumed by standard spring application event listeners.
```

### DIFF
--- a/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/event/TaskReceivedEvent.java
+++ b/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/event/TaskReceivedEvent.java
@@ -1,0 +1,74 @@
+package io.camunda.zeebe.spring.client.event;
+
+import org.springframework.context.ApplicationEvent;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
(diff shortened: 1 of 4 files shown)


berndruecker:
As commented in https://github.com/camunda-community-hub/spring-zeebe/issues/324: You can register job workers via the JobWorkerManager, see how this is done internally after annotations are parsed: https://github.com/camunda-community-hub/spring-zeebe/blob/main/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/annotation/processor/ZeebeWorkerAnnotationProcessor.java#L115

I would not introduce a second way of starting workers - as there is quite some logic involved

Issue #340, opened 2023-02-17
Reconcile identical inbound connectors in case of message events


chillleader:
### Description

Suppose the following diagram:

<img width="863" alt="Screenshot 2023-02-17 at 19 05 39" src="https://user-images.githubusercontent.com/38818382/219742402-0cc2e7f6-8676-40d1-a7a1-70f904b9f15f.png">

There are 2 types of behaviour I may have intended with this diagram:
- **Case 1:** `Payment succeeded` and `Payment failed` events come from different queues, and thus require different subscriptions
- **Case 2:** Both events come from the same queue and only require one subscription

We need to provide the user with a way to distinguish between these types of behaviour and support both cases in the Connector runtime. The current implementation only supports Case 1.

### Why support Case 2?

Consider an example:
I have a message queue with payment results. It can produce 2 types of messages:
1. `{"paymentId": "XXX", "status": "COMPLETED}`
2. `{"paymentId": "XXX", "status": "FAILED}`

If we only support Case 1, with the diagram above I will end up with 2 different consumers. Then I have to additionally worry about the following: How to make sure that each consumer reads only its own messages? This potentially increases the complexity of the Connector implementation and delegates the problem, which can be solved on the runtime level, to the user.

chillleader:
Potential solution: add a new optional property called `Execution Key`
- Connectors with the same execution key will trigger only one subscription. At the same time, other properties of these connectors can be different, including `messageName` or `activationCondition`
- Connectors with different execution keys will trigger separate independent subscriptions
- If no execution key is specified, then this property receives a random value (i.e. separate subscriptions are created)

For those inbound connector types where this problem is irrelevant (for example, webhook, since it does not create any subscriptions) this property can simply be omitted in the element template.

berndruecker:
I like the idea of having a sensible default and being able to overwrite it, if necessary. 
Spontaneously I thought about a property called `subscriptionSharingLabel` or the like - if this is the same - it will be shared. By default, own subscriptions are opened. WDYT?

chillleader:
> Spontaneously I thought about a property called `subscriptionSharingLabel` or the like (but don't like `execution key`) - if this is the same - it will be shared. By default, own subscriptions are opened. WDYT?

How about `connectorInstanceId`? Or simply `deduplicationId` - this one should be intuitively clear without introducing any new concepts. I think that what I mentioned can also be applied to polling connectors, so I would ideally avoid explicitly referencing subscription.

[...]

PR #343, opened 2023-02-21
feat: refine inbound Connectors runtime

chillleader:
- Moved *InboundConnectorContext* implementation to Connector SDK
- Support of Connector SDK changes listed here: https://github.com/camunda/connector-sdk/pull/386
- `InboundConnectorManager` now groups connectors by `execution ID` ([see reasoning](https://github.com/camunda-community-hub/spring-zeebe/issues/340))

___
⚠️ Build depends on https://github.com/camunda/connector-sdk/pull/386
related to https://github.com/camunda-community-hub/spring-zeebe/issues/340
related to https://github.com/camunda/connector-sdk/issues/385


connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorExecutable.java
```diff
@@ -0,0 +1,27 @@
+package io.camunda.connector.runtime.inbound.webhook;
+
+import io.camunda.connector.api.annotation.InboundConnector;
+import io.camunda.connector.api.inbound.InboundConnectorContext;
+import io.camunda.connector.api.inbound.InboundConnectorExecutable;
+import io.camunda.connector.impl.inbound.InboundConnectorProperties;
+
+@InboundConnector(name = "WEBHOOK", type = WebhookConnectorRegistry.TYPE_WEBHOOK)
+public class WebhookConnectorExecutable implements InboundConnectorExecutable {
```
igpetrov, line 9: I am wondering, whether we somehow could move (not in this PR but in another one) this to `connectors-bundle` repository, since all `@OutboundConnector`-s are there.

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConditionalOnProperty("camunda.connector.webhook.enabled")
+public class WebhookConnectorRegistry {
+
+  public static final String TYPE_WEBHOOK = "io.camunda:webhook:1";
```
igpetrov, line 17: I am a bit confused here. You specify type webhook as constant. Will the arbitrary webhooks work?
For example, existing [HTTP](https://github.com/camunda/connectors-bundle/blob/main/connectors/webhook-connector/element-templates/webhook-connector.json#L4) and [GitHub](https://github.com/camunda/connectors-bundle/blob/main/connectors/github/element-templates/github-webhook-connector.json#L4)?

chillleader: Yes, they will: [previously](https://github.com/camunda-community-hub/spring-zeebe/blob/5dd18aeb23c10f29167a12261f6800253c6f17f9/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/registry/InboundConnectorProperties.java#L23) it was also defined as a constant, but it was simply `webhook.` I changed it to match the type pattern that we already have for all outbound connectors.

For custom templates to work, [this property](https://github.com/camunda/connectors-bundle/blob/0655593e5c84223bd2a4ddf257f8e5790a0eafd6/connectors/webhook-connector/element-templates/webhook-connector.json#L32) will need to be changed from `webhook` to `io.camunda:webhook:1`.

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorExecutable.java
chillleader: Currently it's not possible, because `WebhookConnectorExecutable` depends on runtime-specific components [`WebhookConnectorRegistry`](https://github.com/camunda-community-hub/spring-zeebe/blob/5972ea8833eb160fa561829277f0b818556e927c/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java), and transitively [`InboundWebhookRestController`](https://github.com/camunda-community-hub/spring-zeebe/blob/5972ea8833eb160fa561829277f0b818556e927c/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java), which can only exist in Runtime.

I think it may be possible later on, but not sure if it makes sense - since the actual endpoint code will always live in the runtime, the executable can't be used by itself, separately from the runtime.

igpetrov (approved):

PR #348, opened 2023-02-24
feat(): catch and throw exceptions too

### LINKED ISSUE #347: `io.camunda.zeebe.spring.client.bean.MethodInfo#invoke` wraps all non-runtime exceptions
I think that we can change this condition a bit, and throw the `java.lang.Exception` class without worrying because in the [`JobHandler`](https://github.com/camunda/zeebe/blob/6108195a9c5f443a8d4aa62ba556cb0fe668278d/clients/java/src/main/java/io/camunda/zeebe/client/api/worker/JobHandler.java#L35) it is already can throw `java.lang.Exception`.

aivinog1:
I modified the catch-throw block, so we could throw not only `RuntimeException`s but `Exception`s too

Closes #347

### DIFF
--- a/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/MethodInfo.java
+++ b/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/bean/MethodInfo.java
@@ -41,13 +41,13 @@ public class MethodInfo implements BeanInfo {
     return method.getName();
   }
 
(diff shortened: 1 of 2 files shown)


berndruecker:
I can't think of any concrete problem this causes - just that it changes behavior in a non backwards compatible way (e.g. because some users rely on unwrapping the existing RuntimeException somewhere). Happy to merge it.

Issue #342, opened 2023-02-21
Add an endpoint to return information about active webhooks for a diagram


nithinssabu:
### Context

We are tyring to enhance the UX for users modeling inbound webhook connectors in Web Modeler by [providing the Webhook URL ](https://github.com/camunda/web-modeler/issues/3335)the users can easily copy. Although the URL can be computed in Web Modeler itself using the following format,

`https://ConnectorURL/clusterId/inbound/webhookId`

Web Modeler doesn't know if the URL is really active.

### Description

Add an endpoint that will return the information about active webhooks for a given diagram based on the provided diagram id.

### Proposed solution

Web Modeler passes the diagram id to the endpoint and the responsse can be:

- Webhook ID, if the Webhook is active
- Empty, if the Webhook is not active

chillleader:
Hi @nithinssabu, I drafted the API contract for this endpoint. Please let me know what you think about this.

### Request 
GET `${connectorsBaseUrl}/inbound`

Query parameters:
- `bpmnProcessId` - diagram id (optional parameter)
- `type` - connector type (optional parameter).

`type` parameter can be used to filter the running inbound connectors by type. Currently, we will only have one type of inbound connectors - `io.camunda:webhook:1`. But this will soon change, when we add support for e.g. RabbitMQ subscription or HTTP polling, which will require a similar UI feature in the Modeler.
Connector's `type` can be found in the element template JSON, and also in the process XML (but not sure if you have convenient access to it in the Modeler).

If no parameters are passed, then all active connectors will be returned - for all diagrams and all connector types.

### Response example

```
[
  {
    "bpmnProcessId": "8b1a3ac7-2edd-4d54-9c65-aa983c65edd8",
    "type": "io.camunda:webhook:1",
    "data": {
      "path": "/my-inbound-connector-path"
    },
    {
    "bpmnProcessId": "8b1a3ac7-2edd-4d54-9c65-aa983c65edd8",
    "type": "io.camunda:webhook:1",
    "data": {
      "path": "/my-other-inbound-connector-path"
    }
  }
]
```

nithinssabu:
> Hi @nithinssabu, I drafted the API contract for this endpoint. Please let me know what you think about this.

Thanks for setting this up quickly, @chillleader. It looks good to me overall.

I have two things to clarify though:

1. The first parameter name is called `bpmnProcessId` but you expect the diagram id. Diagram id is what we see in the URL 
[...]

PR #352, opened 2023-02-28
feat(connectors): add endpoint to retrieve active inbound connectors

chillleader:
Adds an endpoint to retrieve all active inbound connectors.

**Sample response structure:**
```
[
  {
    "bpmnProcessId": "8b1a3ac7-2edd-4d54-9c65-aa983c65edd8",
    "type": "io.camunda:webhook:1",
    "data": {
      "path": "my-inbound-connector-path"
    },
    {
    "bpmnProcessId": "8b1a3ac7-2edd-4d54-9c65-aa983c65edd8",
    "type": "io.camunda:webhook:1",
    "data": {
      "path": "my-other-inbound-connector-path"
    }
  }
]
```

**Current limitations:**
In Spring Zeebe runtime, we only know about the internals of the Webhook Connector. The Runtime can't possibly know what data should be displayed in the Modeler for generic inbound connectors (e.g. various subscriptions).

We will need to evolve this into a more robust and generic solution, that would allow Connector developers to indicate what information about the inbound connector execution should be displayed in the Modeler.

related to #342

### DIFF
--- a/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
+++ b/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
@@ -1,0 +1,71 @@
+package io.camunda.connector.runtime.inbound.lifecycle;
+
+import io.camunda.connector.impl.inbound.InboundConnectorProperties;
+import io.camunda.connector.runtime.inbound.webhook.WebhookConnectorProperties;
+import io.camunda.connector.runtime.inbound.webhook.WebhookConnectorRegistry;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
```diff
@@ -0,0 +1,71 @@
...
+    Map<String, Set<InboundConnectorProperties>> connectors, String bpmnProcessId) {
+    if (bpmnProcessId != null) {
+      return new ArrayList<>(connectors.get(bpmnProcessId));
+    }
+    return connectors.values().stream()
+      .flatMap(Collection::stream)
+      .collect(Collectors.toList());
+  }
+
+  private List<InboundConnectorProperties> filterByType(
```
igpetrov, line 54: would rather suggest using more explicit names. Maybe something like `filterByConnectorType`. Same for `extractData`, I would suggest to give more explicit name.

igpetrov (approved):

Issue #367, opened 2023-03-09
Prepare spring-zeebe to evolve towards a "camunda-8-spring-sdk"


berndruecker:
We want to develop spring-zeebe towards being the Camunda 8 Spring SDK (like the Node SDK: https://github.com/camunda-community-hub/camunda-8-sdk-node-js). Camunda-internal doc: https://docs.google.com/document/d/1W7VueAdUJgQoGxY1X_H_DJimSEX1hxT-JJjKfaBzrFY/edit

As a first step we can prepare the spring-zeebe project for that move:

- [x] Make sure all “Zeebe Client” logic is in its own sub module (based on https://github.com/camunda-community-hub/spring-zeebe/tree/main/client/spring-zeebe)
- [x] Add own sub module for other clients, starting with Operate, somehow based on [existing community extension](https://github.com/camunda-community-hub/camunda-operate-client-java)
- [x] Separate the “starter” application (Spring Auto Configuration) where you can configure everything and switch components on/off (e.g. you only want to connect to Zeebe, not Operate - or the other way round)
- [x] Extract connector runtime (https://github.com/camunda-community-hub/spring-zeebe/tree/main/connector-runtime) and move it to an own place (probably within https://github.com/camunda/connectors-bundle)
- [x] Add legacy warnings to old artifact names
- [x] Double check naming of configuration options
- [x] Adjust documentation and write update docs

Next steps will be:
- Rename Spring-Zeebe to Camunda 8 Spring SDK
- Add clients for Tasklist, Optimize, Console

berndruecker:
Thinking about naming (of course) - current proposal: 

* Package: `io.camunda.zeebe.spring` (I would not touch this, as this would require too much changes for existing users)
* Maven GroupId: `io.camunda.spring` (I would adjust now as we are cleaning up and forcing people to adjust dependencies anyway)
* Maven ArtifactIds:
  * spring-base / common / util / ...? I think we will move some things out of zeebe-client (e.g. Metrics) as this could also be used for operate-client. Just naming 🤔 
  * spring-zeebe-client
  * spring-operate-client
  * spring-boot-starter-camunda
  * spring-boot-starter-camunda-test-common
  * spring-boot-starter-camunda-test
  * spring-boot-starter-camunda-test-testcontainer

We could keep some legacy dependencies:
```
  <groupId>io.camunda</groupId>
  <artifactId>spring-zeebe-starter</artifactId>
```
would simply pull in
```
[...]

Issue #370, opened 2023-03-13
Inconsistent webhook endpoint behavior after disabling


chillleader:
Webhook endpoint responds inconsistently in similar situations.

**To reproduce:**
- Run a connectors deployment from scratch
- Send a request to the webhook endpoint with random inexistent path, **ensure 404 is returned**
- Deploy a process with webhook start event connector
- Deploy a newer version without webhook (thus disabling it)
- **Ensure 200 is returned with empty response**

**Expected behavior**

We shouldn't differentiate between cases when webhook wasn't. enabled and when it was disabled. We should return 404 consistently.

PR #372, opened 2023-03-14
Moved Operate Client to Starter module and cleaned up AutoConfiguration

berndruecker:
…ons furtherRelated tp https://github.com/camunda-community-hub/spring-zeebe/issues/367


berndruecker:
@chillleader Do you plan another alpha with the Metrics fix soon? Then I might hold this back a little bit - otherwise I would go ahead and also merge this PR that moved the OperateClient to the starter (and thus removed it from the ConnectorRuntime)

chillleader:
> @chillleader Do you plan another alpha with the Metrics fix soon? Then I might hold this back a little bit - otherwise I would go ahead and also merge this PR that moved the OperateClient to the starter (and thus removed it from the ConnectorRuntime)

I was thinking of maybe doing another alpha release tomorrow. I'd like to also include https://github.com/camunda-community-hub/spring-zeebe/issues/370.

chillleader:
@berndruecker FYI: I did an alpha2 release, so you can go ahead with merging this now 🙂

PR #375, opened 2023-03-21
add missing autoComplete = false

ingorichtsmeier:
The example code looks wrong, as it overrides the default behavior of autoComplete.


berndruecker:
Good catch 👍

PR #383, opened 2023-03-28
fix: `ZeebeClientConfigurationProperties` doesn't work with Spring Boot 3

### LINKED ISSUE #382: `ZeebeClientConfigurationProperties` is not initialized correctly when using Spring Boot 3
When using spring-boot-starter-camudna with Spring Boot 3, configuration properties for Zeebe client are not instantiated correctly.

### How to reproduce:
1. Include the starter dependency in your spring boot 3.0.x-based project
2. Run the application with at least one job worker configured
3. Observe the following (or similar) error messages:

```
2023-03-28T13:26:35.363+02:00  WARN 52401 --- [ult-executor-13] io.camunda.zeebe.client.job.poller       : Failed to activate jobs for worker SLACK and job type io.camunda:slack:1

io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
...
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: /0.0.0.0:26500
Caused by: java.net.ConnectException: Connection refused
```

### Analysis

The issue is caused by [this change](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M2-Release-Notes#improved-constructorbinding-detection) in Spring Boot 3.
Following the deprecation of `@ConstructorBinding`, any `ConfigurationProperties` class with a constructor will only be instantiated using its constructor (and setters are ignored), which breaks `ZeebeClientConfigurationProperties`, where injections both via constructor and via setters are expected.

The solution would be to mark the constructor of `ZeebeClientConfigurationProperties` with `@Autowired` to prevent it being treated as a `@ConstructorBinding` configuration properties.

chillleader:
### Description

Moved `@Autowired` to the constructor level to comply with the [changes in Spring Boot 3](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M2-Release-Notes#improved-constructorbinding-detection).
See also the related issue for more context.

### Related issues
closes #382

### DIFF
--- a/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
+++ b/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
@@ -74,7 +74,8 @@ public class ZeebeClientConfigurationProperties implements ZeebeClientConfigurat
   @Autowired(required = false)
   private List<ClientInterceptor> interceptors;
 
-  public ZeebeClientConfigurationProperties(@Autowired org.springframework.core.env.Environment environment) {
+  @Autowired
+  public ZeebeClientConfigurationProperties(org.springframework.core.env.Environment environment) {
     this.environment = environment;
   }
 


markfarkas-camunda (approved):
Thank you! Nice fix 🥇

PR #391, opened 2023-04-01
325 i can disable connectors via config

l646505418:
1. enhancement for disabling the whole connectors by adding
`zeebe.client.connector.enabled=false` 
and disabling specific connector by adding
`zeebe.client.connector.properties.Twitter.enabled=false`
ps:`Twitter `is connector's name

2.Add a interface `ZeebeConnectorFilter` and users can implement it to customize their disable rule


spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ConnectorConfiguration.java
```diff
@@ -5,34 +5,33 @@
 import io.camunda.connector.impl.config.ConnectorPropertyResolver;
 import io.camunda.connector.runtime.util.outbound.DefaultOutboundConnectorFactory;
 import io.camunda.connector.runtime.util.outbound.OutboundConnectorFactory;
-import io.camunda.zeebe.spring.client.connector.OutboundConnectorManager;
+import io.camunda.zeebe.spring.client.annotation.customizer.ZeebeWorkerValueCustomizer;
+import io.camunda.zeebe.spring.client.annotation.processor.AnnotationProcessorConfiguration;import io.camunda.zeebe.spring.client.connector.Filter.ZeebeConnectorFilter;import io.camunda.zeebe.spring.client.connector.OutboundConnectorManager;
```
chillleader, line 9: Please reformat the code to match the project code style, specifically to avoid multiple imports in one line

chillleader (commented):
Hi @l646505418 (cc @berndruecker),

All functionality related to Connectors is being moved to [another repository](https://github.com/camunda/connectors-bundle). The code changed in this PR is also affected.

I'm sorry to see that this PR got stuck - let's fix it! Here are 2 options of what we can do:

1. We can merge it as soon as possible, and I will transfer it together with the rest of the code. Please note that in that case I cannot guarantee that the exact commit history will be preserved, but I will at least make sure that @l646505418 is listed as a contributor.
2. Alternatively, we can close this PR and move the Connectors code without these changes. In that case, @l646505418 you can open a new PR in the other repository. In that case the implementation will need to be adjusted - but we can also review it really quickly in the connectors-bundle repo.

Please let me know what you would prefer, or if you have any other thoughts/concerns.

berndruecker:
@chillleader and @l646505418: Sorry for the delay on this on my end - first vacations and now other priorities - so I could not have a look at it again. From my perspective this PR should not block any Move-Effort and I think it would be not too hard to re-apply in the new repo (as I will not find to react to the raised discussions this week). 
@l646505418: WDYT?

sbuettner:
Closing this as the issue was move to https://github.com/camunda/connectors/issues/518

Issue #401, opened 2023-04-05
Store inbound connector errors


chillleader:
### What do we need to do?

Connector runtime needs to store errors that happen in the following cases:
- During inbound connector activation (i.e. in a synchronous call to `activate()` method)
- During `InboundConnectorExecutable` lifecycle (i.e. when connector declares an error by calling the `cancel()` method of the `InboundConnectorContext`.

We also need to return this information from our status endpoint (GET `/inboung`) alongside running connectors.

### Why do we need to do it?

To provide the user with debug information about errors that happened during inbound connector execution.

chillleader:
This issue was transferred to [connectors-bundle](https://github.com/camunda/connectors-bundle/issues/514) repo.

PR #402, opened 2023-04-05
feat(connectors): bump connector-sdk

chillleader:
Adjustments to Connector runtime to support the [changes in the Connector SDK](https://github.com/camunda/connector-sdk/pull/411):

- Variable mapping and activation condition check are now handled centrally in the SDK (added deprecation warning to legacy checks in webhook connector)
- Adjustments to support slightly changed APIs
- Cancellation callback handling (initial simple implementation, just deregisters the connector without storing the error). Will be improved as a follow up: #401

### DIFF
--- a/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
+++ b/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
@@ -123,6 +123,7 @@ public class InboundConnectorManager {
         "Inbound connector failed at correlation point " + newProperties.getCorrelationPointId(),
         throwable);
       // TODO: store error for user's convenience
+      // (see https://github.com/camunda-community-hub/spring-zeebe/issues/401)
       deactivateConnector(newProperties);
     };
 
--- a/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
+++ b/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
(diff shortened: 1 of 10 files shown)


connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
```diff
@@ -117,10 +118,18 @@ private void deactivateConnector(InboundConnectorProperties properties) {
   private void activateConnector(InboundConnectorProperties newProperties) {
 
     InboundConnectorExecutable executable = connectorFactory.getInstance(newProperties.getType());
+    Consumer<Throwable> cancellationCallback = throwable -> {
+      LOG.error(
+        "Inbound connector failed at correlation point " + newProperties.getCorrelationPointId(),
+        throwable);
+      // TODO: store error for user's convenience
```
igpetrov, line 125: It would be great if we can convert all todos under one or several tickets to visualize scope for the future work and tracking.

connector-runtime/src/main/resources/application.properties
```diff
@@ -3,7 +3,6 @@ camunda.connector.polling.enabled=false
 camunda.connector.polling.interval=5000
 
 camunda.connector.webhook.enabled=false
-spring.main.web-application-type=none
```
igpetrov, line 6: Wouldn't that always activate MVC REST endpoints?

igpetrov (approved):
LGTM. Few questions

connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
chillleader: This one should cover the todos at least in this class: https://github.com/camunda-community-hub/spring-zeebe/issues/401

connector-runtime/src/main/resources/application.properties
chillleader: Restored - let's keep it disabled for now, I will test later if we really need this switch here since we have a few endpoints in the Runtime now.

PR #427, opened 2023-05-10
Updated README.md to match #275

mKrswe:
Deleted unnecessary advice to use `@EnableZeebeClient` for enabling the ZeebeClient.
The annotation is deprecated after implementation of #275.


berndruecker:
Thank you @TakePack !

PR #449, opened 2023-06-27
Changed examples and README to match reality

marcosdiez:
This costed about 5 hours of work from 7 people yesterday. I hope others can benefit from our wasted time.


berndruecker:
Thanks @marcosdiez for adjusting this! I think Spring should be able to read both ways of config properties - but happy to get this into the docs (and haven't double checked)

daniel-ewing:
Hi @berndruecker, I think it must be a limitation for application.properties. kebab-case for these properties works in my yaml file.

PR #455, opened 2023-07-11
Fix issue for keycloak17

chDame:
bring the possibility to force keycloak token url rather than building it from the realm and the base url


igpetrov (approved):
LGTM, thank you.

Maybe a suggestion to do some reformatting so that the code looks in the same style.

PR #457, opened 2023-07-17
Update to current artifact, replacing the deprecated coordindates

tobiasschaefer:
Please double check: this documentation seems out-dated because the artifact was moved to `io.camunda.spring:spring-boot-starter-camunda`


berndruecker:
Thanks a lot @tobiasschaefer

PR #459, opened 2023-07-26
bugfix OperateClientConfigurationProperties.java

chDame:
keycloak url is provided as a default from the helm charts and cannot easily be unset. So we should change the test order. If the keycloak token url is provided we use that, else we do the default.


chDame:
Hi team, as the helm charts provide keycloak url as a default environment variable, our prospect is not able to override it with the token url. Thanks in advance for your review and pinging me when a new release of spring zeebe client and connector runtime is available :)

berndruecker:
Sounds good to me - do you think you can add a comment in the source code adding the info why the sequence is like it is? I think this would be valuable for possible future changes

berndruecker (approved):

berndruecker:
For the release date: Is this urgent @chDame ? I could issue a release with this change if helpful - there is no fixed cadence for spring-zeebe at the moment

chDame:
> For the release date: Is this urgent @chDame ? I could issue a release with this change if helpful - there is no fixed cadence for spring-zeebe at the moment

Hi @berndruecker, I have a prospect waiting for it. Would be nice if we can have a release soon :) Thanks!

berndruecker:
I will release it now

berndruecker:
Released @chDame : https://github.com/camunda-community-hub/spring-zeebe/releases/tag/8.2.4

chDame:
Thanks Bernd!

PR #462, opened 2023-08-09
Fixes #461

### LINKED ISSUE #461: NullPointerException in some case
io.camunda.zeebe.spring.client.bean.MethodInfo#getParameters and io.camunda.zeebe.spring.client.bean.MethodInfo#getParametersFilteredByAnnotation methods can throw a NullPointerException when org.springframework.core.LocalVariableTableParameterNameDiscoverer#getParameterNames(java.lang.reflect.Method) returns null

wallaroo:
nullpointer protection when building native-image

Fixes #461

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,14 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>3.1.1</version>
+        <configuration>
+          <skip>false</skip>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 
@@ -295,4 +303,13 @@
     <developerConnection>scm:git:git@github.com:camunda-community-hub/spring-zeebe.git</developerConnection>
     <tag>HEAD</tag>
   </scm>
+  <distributionManagement>
+    <snapshotRepository>
+      <id>a4b-multiverse</id>
+      <url>https://gitlab.appetite4byte.com/api/v4/projects/78/packages/maven</url>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </snapshotRepository>
+  </distributionManagement>
 </project>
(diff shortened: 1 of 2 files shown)


berndruecker:
@wallaroo: Could you please remove the Maven build changes (in pom.xml) from the PR - they are unrelated from the issue and actually not intended for us. Thx

PieterSchutte:
I dont think this fix will address the root issue. In our case, the parameters that are causing the null is the JobClient and the ActivatedJob. I think the fix outlined will result in those parameters being null. We also have this issue only appearing in our cloud environment and not on our local environments. (Camunda 8.2.5 and spring client 8.2.4)

1nb0und:
@wallaroo @PieterSchutte Can you provide more insight on what was the load like in `spring-zeebe` when this issue happens (so that we can replicate this) ? i.e. does this happen when there are parallel tasks, or serially.

Would still need more time to triage this issue. Maybe we can follow up with a patch release once a fix has been created.

wallaroo:
@berndruecker yep sorry for the laggy response...
@1nb0und i'm pretty sure its 'cause i was building a spring native image (GraalVM) and i wasn't able to address this issue using reflectconfig.

with classic image (JVM) there's no problem with this

PieterSchutte:
@1nb0und  There was no load (dev env with a couple of processes and about 10 process instances), no parallel tasks, works flawlessly when connected from intellij. Once Microservice is deployed in Pivotal Cloud Foundry, the issue occurred for every job triggered, on every jobworker. Apologies I do not have more info

1nb0und:
According to the latest suggestion and comments in the corresponding [issue](https://github.com/camunda-community-hub/spring-zeebe/issues/461) , the null pointer is caused by missing java compiler `-parameter` flag. Once that flag is enabled, then variable should no longer be null. [Readme](https://github.com/camunda-community-hub/spring-zeebe?tab=readme-ov-file#add-spring-boot-starter-to-your-project) also reflects how to use the new parameter

Closing this. Let's reopen another issue @wallaroo if this isn't the case for you.

PR #469, opened 2023-09-04
chore: Upgrade dependencies for multi-tenancy alpha support

sbuettner:
This upgrades the dependencies for the zeebe and operate clients to the latest alpha versions to allow the connectors component to make use of multi-tenancy. This also includes an update of the jackson dependency as the previous pom did not provide an up-to-date version that was compatible with the zeebe client.

### DIFF
--- a/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
+++ b/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
@@ -47,6 +47,10 @@ public class ZeebeClientConfigurationProperties implements ZeebeClientConfigurat
    */
   private String connectionMode;
 
+  private String defaultTenantId = DEFAULT.getDefaultTenantId();
+
+  private List<String> defaultJobWorkerTenantIds = DEFAULT.getDefaultJobWorkerTenantIds();
+
   private boolean applyEnvironmentVariableOverrides = false; // the default is NOT to overwrite anything by environment variables in a Spring Boot world - it is unintuitive
 
   private boolean enabled = true;
@@ -82,6 +86,8 @@ public class ZeebeClientConfigurationProperties implements ZeebeClientConfigurat
   private ScheduledExecutorService scheduledExecutorService;
 
   private boolean ownsJobWorkerExecutor;
+
+  private boolean defaultJobWorkerStreamEnabled = DEFAULT.getDefaultJobWorkerStreamEnabled();
 
   @Autowired
   public ZeebeClientConfigurationProperties(org.springframework.core.env.Environment environment) {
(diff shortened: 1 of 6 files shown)


chillleader (approved):

chillleader (approved):

sbuettner:
@berndruecker How should we proceed with PRs that target multi-tenancy? Wait until its officially released in dependencies?

chillleader (approved):

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
```diff
@@ -194,6 +201,24 @@ public void setApplyEnvironmentVariableOverrides(boolean applyEnvironmentVariabl
...
+  public void setScheduledExecutorService(ScheduledExecutorService scheduledExecutorService) {
+    this.scheduledExecutorService = scheduledExecutorService;
+  }
+
+  public void setOwnsJobWorkerExecutor(boolean ownsJobWorkerExecutor) {
+    this.ownsJobWorkerExecutor = ownsJobWorkerExecutor;
+  }
+
+  @Override
+  public ScheduledExecutorService jobWorkerExecutor() {
```
1nb0und, line 213:
```suggestion
  public ScheduledExecutorService getJobWorkerExecutor() {
```

Whats your thought about prefixing this with `get` similar to other getters?

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
```diff
@@ -194,6 +201,24 @@ public void setApplyEnvironmentVariableOverrides(boolean applyEnvironmentVariabl
...
+    return scheduledExecutorService;
+  }
+
+  @Override
+  public boolean ownsJobWorkerExecutor() {
```
1nb0und, line 218: Same as above
```suggestion
  public boolean getOwnsJobWorkerExecutor() {
```

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
sbuettner: @1nb0und Good idea but as it overrides an interface implementation we cant just rename it here. Its coming from: https://github.com/camunda/zeebe/blob/b1a13e14494322053dd6d42f477854357ed5ec3c/clients/java/src/main/java/io/camunda/zeebe/client/ZeebeClientConfiguration.java#L110

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
sbuettner: Please see other related comment above.

PR #476, opened 2023-10-03
chore(docs): update version matrix

1nb0und:
The previous matrix is okay but might be outdated. Wanted to see if an expanded version will be better. The old matrix is simpler but might omit some tiny details. The new matrix includes the following:

- JDK scope was included till 8 (not sure if its worth to include 8 given its End of Life already)
- Added separate rows for JDK 17 (only required for SB 3.x)


berndruecker (approved):
LGTM 👍

Issue #472, opened 2023-09-12
Check if Java Client Micrometer Metrics are picked up by spring-zeebe


berndruecker:
As I just learned the Java Client has added some metrics. We should check if that is pulled into spring-zeebe correctly

npepinpe:
It has to be explicitly enabled, since it's an optional dependency. e.g.

```java
final PrometheusMeterRegistry prometheusRegistry = ...;
final JobWorkerMetrics metrics =
    JobWorkerMetrics.micrometer()
        .withMeterRegistry(prometheusRegistry)
        .withTags(Tags.of("workerName", "myCoolWorker", "jobType", jobType))
        .build();
```

1nb0und:
Here is a first cut [PR](https://github.com/camunda-community-hub/spring-zeebe/pull/486) for enabling java-client metrics

1nb0und:
Moving this to the next milestone after receiving feedback

1nb0und:
Metrics PR has been merged

PR #486, opened 2023-10-10
Enable Java Client / JobWorker Metrics

1nb0und:
### Overview
Zeebe's java-client has JobWorker [metrics](https://github.com/camunda-community-hub/spring-zeebe/issues/472) and explicitly enabling it on `spring-zeebe`

Zeebe-client metrics gets translated towards Spring's `MetricRecorder` so that users will see consistent metrics in the format of `metric` and `action`.

### Changes
- enabled by default (can be disabled by config property if there is a need?)
- users will see 2 sets of metrics in the same format:
  - `camunda.job.invocations` (from spring)
  - `zeebe.client.worker.job` (from zeebe-client)

### DIFF
--- a/spring-client-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobWorkerManager.java
+++ b/spring-client-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobWorkerManager.java
@@ -70,7 +70,7 @@ public class JobWorkerManager {
       builder.fetchVariables(zeebeWorkerValue.getFetchVariables());
     }
 
-    // TODO: Check if this is the right place for enabling JobWorker metrics
+    // TODO: Check if this is the right place for enabling JobWorker metrics and connect to Meter Registry
     try {
       JobWorkerMetrics metrics = JobWorkerMetrics.micrometer().build();
       builder.metrics(metrics);
--- a/spring-client-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobWorkerManager.java
+++ b/spring-client-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobWorkerManager.java
@@ -5,9 +5,9 @@ import io.camunda.zeebe.client.api.JsonMapper;
 import io.camunda.zeebe.client.api.worker.JobHandler;
 import io.camunda.zeebe.client.api.worker.JobWorker;
 import io.camunda.zeebe.client.api.worker.JobWorkerBuilderStep1;
-import io.camunda.zeebe.client.api.worker.JobWorkerMetrics;
 import io.camunda.zeebe.spring.client.annotation.value.ZeebeWorkerValue;
 import io.camunda.zeebe.spring.client.metrics.MetricsRecorder;
+import io.camunda.zeebe.spring.client.metrics.ZeebeClientMetricsBridge;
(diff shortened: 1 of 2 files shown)


berndruecker:
Couple of thoughts:

1. I think we should use the MetricsRecorder provided by Spring Boot which is injected into this class in `MeterRegistry` (see MicrometerMetricsRecorder)
2. It looks like the Java Client added some magic around building their own MetricsRecorder, but only if Micrometer is on the classpath. I think we should avoid using this in the Spring environment alltogether.

I think therefore me might need to copy the creation of the JobWorkerMetrics class, as this seems to be required:

https://github.com/camunda/zeebe/blob/503ae8e5da9cade1a9471e436051542a0a78d0ee/clients/java/src/main/java/io/camunda/zeebe/client/impl/worker/metrics/MicrometerJobWorkerMetricsBuilderImpl.java#L44-47

In essence, this is a good example, where we replicate Spring magic in the Java client, which even can make some trouble in the spring integration now (but haven't looked at all the details).

Does this make sense?

Then I think we should adjust this - so that all Spring-zeebe uses the MeterRegistry from Spring (as people could create their own if they want - then this should be used - to stick to normal Spring principles). WDYT @1nb0und ?

1nb0und:
> Then I think we should adjust this - so that all Spring-zeebe uses the MeterRegistry from Spring (as people could create their own if they want - then this should be used - to stick to normal Spring principles). WDYT @1nb0und ?

Those are good points. It makes sense to tap into the existing metrics recorder rather. Will adjust.

npepinpe:
Please let us know how we can make the integration with Spring easier for you. As `MetricsRecorder` seems to be `spring-zeebe` specific, maybe instead you can simply create an adapter which bridges that and `JobWorkerMetrics`?

Something like:

```java
public final class MetricsRecordJobWorkerMetricsAdapter implements JobWorkerMetrics {
  private final MetricsRecorder delegate;

  public MetricsRecordJobWorkerMetricsAdapter(final MetricsRecorder delegate) {
    this.delegate = delegate;
  }

  @Override
  public void jobActivated(final int count) {
    delegate.increase("zeebe.client.worker.job", "activated", count);
  }

  @Override
  public void jobHandled(final int count) {
    delegate.increase("zeebe.client.worker.job", "handled", count);
  }
}
```

You can then use it:

```java
client
  .newWorker()
  .type("type")
  .handler(...)
  .metrics(new MetricsRecordJobWorkerMetricsAdapter(metricsRecorder))
  .open();
```

Our metric names are actually part of the public API (`MicrometerJobWorkerMetricsBuilder.Names`), but unfortunately we don't split between the metric and the action, so it wouldn't work out of the box with the `MetricsRecord` API.

1nb0und:
Ok I'm understanding the problem space now. What was happening previously was since zeebe-client didnt have any metrics implementation, it was `spring-zeebe` who provided it via the `MetricsRecorder`. Now that the `zeebe-client` implemented it, I would need to reconcile the two and ensure its done in a clean manner.

Thank you both for your insights.

1nb0und:
We are now passing metrics from `zeebe-client` to `spring-zeebe` thru a bridge class. Also this ensures the creation of metrics in Spring will be in a single place at `MetricsRecorder`. 

Looking in the future for multiple clients like Operate and TaskList, if they have metrics to be exposed, then there will be multiple bridges/adapters which all gets translated in `MetricsRecorder`

Issue #499, opened 2023-10-25
Resilience4j dependency pulls in a transitive slf4j which is too old


berndruecker:
Running a plain worker (like https://github.com/camunda/camunda-platform-tutorials/tree/main/orchestrate-microservices/worker-java) on spring-zeebe 8.2.x or 8.3.0 will result in the following exception:

```log
[WARNING]
java.lang.IllegalArgumentException: LoggerFactory is not a Logback LoggerContext but Logback is on the classpath. Either remove Logback or the competing implementation (class org.slf4j.helpers.NOPLoggerFactory loaded from file:/C:/Users/ruecker/.m2/repository/org/slf4j/slf4j-api/1.7.30/slf4j-api-1.7.30.jar). If you are using WebLogic you will need to add 'org.slf4j' to prefer-application-packages in WEB-INF/weblogic.xml: org.slf4j.helpers.NOPLoggerFactory
    at org.springframework.util.Assert.instanceCheckFailed (Assert.java:713)
    at org.springframework.util.Assert.isInstanceOf (Assert.java:632)
    at org.springframework.boot.logging.logback.LogbackLoggingSystem.getLoggerContext (LogbackLoggingSystem.java:389)
    at org.springframework.boot.logging.logback.LogbackLoggingSystem.beforeInitialize (LogbackLoggingSystem.java:123)
    at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationStartingEvent (LoggingApplicationListener.java:238)
    at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationEvent (LoggingApplicationListener.java:220)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener (SimpleApplicationEventMulticaster.java:174)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener (SimpleApplicationEventMulticaster.java:167)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent (SimpleApplicationEventMulticaster.java:145)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent (SimpleApplicationEventMulticaster.java:133)
    at org.springframework.boot.context.event.EventPublishingRunListener.multicastInitialEvent (EventPublishingRunListener.java:136)
    at org.springframework.boot.context.event.EventPublishingRunListener.starting (EventPublishingRunListener.java:75)
[...]

PR #500, opened 2023-10-25
Added slf4j dependency to enforce correct version on transitive dependency

berndruecker:
Solves #499


berndruecker:
@1nb0und Let me know what you think. If we make progress here we should either release a 8.3.1 und update the tutorial (my preference) or add the slf4j dependency also to the tutorial until released

1nb0und:
I prefer this approach and release a 8.3.1 and update the preference so theres less impact for new users trying it out.

I'll merge this now and update the tutorial

berndruecker:
Awesome 👍

1nb0und (approved):

PR #505, opened 2023-10-31
prepare for more examples

ingorichtsmeier:
Moving the one existing example into a subfolder to create space for more examples.


berndruecker:
I am not sure if we want to have more examples directly within Spring Zeebe - normally examples live outside of the core framework. So I think examples should basically end up here: https://github.com/camunda-community-hub/camunda-8-examples. Having one condensed example in Spring ZEebe was an exception, also because of the self-contained community extension setup.

WDYT @1nb0und ?

ingorichtsmeier:
I'm also not sure. 

I have received a library from 1&1 that implements the `io.camunda.zeebe.spring.client.annotation.customizer.ZeebeWorkerValueCustomizer` to allow adjustments of the worker type with Spring Boot configuration. The library helps to create a namespace around the workers in an application.

My initial idea was to place it as an example here.

But the library could live somewhere else as well, perhaps in its own repository on the camunda-community-hub.

What do you think? 

I'm also looking for a good name. The original name was "TIF Zeebe Worker SpringBoot Starter", where TIF is the name of the company department. My first proposal is "dynamic-job-worker-spring-boot-starter". Do you have better names?

ingorichtsmeier:
The idea is the same, but 1&1 have put a lot more effort into configuration options. 

It became a library on top of the spring-boot-starter-camunda. You can apply the configurable namespace on call activities as well.

After the first thoughts, I tend to publish it in the community-hub as its own repo.

ingorichtsmeier:
yes

1nb0und:
I prefer all examples to be in a separate repo, that way a user can have access to different types of example which aids their exploration.

PR #510, opened 2023-11-07
Operate and Commons module

1nb0und:
## Overview
As we transition away from Spring Zeebe to **Camunda 8 Spring SDK**, we need to start supporting products. `Operate` was identified as a first step since `spring-zeebe` has `zeebe` and `operate` intertwined. This effort would also give rise to `commons` module that will handle cross-cutting concerns such as making http calls, authentication, etc (to name a few)

## Disclaimer
This PR is still in **draft** mode but I will continue working on it to address the following:
- Implement Simple Authentication
- Implement the rest of Operate operations (Search, List, XML definition, POST resources)
- More internal testing 
- Will add unit / integration tests once I get feedback and say we are in the right direction

The changes has been done to be backwards compatible as much as possible.

## Changes
- extracted `operate` as its own submodule
- created `commons` module to handle cross-cutting concerns
- abstracted away `Authentication` so that it can be reused across multiple products

## Credit
I'd like to give credit to @chDame as most of the operate logic and classes has been copy pasted from the operate-java-client and further improve upon.

## Open questions and ideas

1. Currently config properties (application.properties) are heavily based on `zeebe`. Can we start making the props for `zeebe` under `camunda` prefix such that:

`camunda.zeebe.x`
`camunda.operate.x`
`camunda.tasklist.x`
... and so on

Reasons are, making `zeebe`, `operate` parent means it might have conflict with other libraries our users will be using. There could be a company, or other library that uses `tasklist` as parent prefix so I think namespacing them to `camunda` might make sense.

2. I'd like to have a commons prefix such that common credentials will be pulled from. User can specify 1 common credential that has access to all products so it might be good to specify a common property instead of declaring them repeatedly

`camunda.common.client.clientId`
`camunda.common.client.clientSecret`

### DIFF
--- a/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/CommonClientConfigurationProperties.java
+++ b/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/CommonClientConfigurationProperties.java
@@ -1,0 +1,146 @@
+package io.camunda.zeebe.spring.client.properties;
+
+import io.camunda.commons.auth.*;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties
+public class CommonClientConfigurationProperties {
+
+  @Value("${zeebe.client.cloud.cluster-id:#{null}}")
+  private String clusterId;
+
+  //@Value("${zeebe.client.cloud.region:bru-2}")
+  //private String region;
+
+  @Value("${zeebe.client.cloud.client-id:#{null}}")
+  private String zeebeClientId;
+
+  @Value("${zeebe.client.cloud.client-secret:#{null}}")
+  private String zeebeClientSecret;
+
+  @Value("${zeebe.client.enabled:#{true}}")
+  private Boolean zeebeEnabled;
+
+  @Value("${zeebe.client.cloud.authUrl:#{null}}")
+  private String authUrlZeebe;
+
+  @Value("${zeebe.client.cloud.baseUrl:#{null}}")
+  private String audienceZeebe;
+
+  // TODO: refactor to its own client props maybe?
+  // operate - check why zeebe has different casing with dash '-'
+  @Value("${camunda.operate.client.clientId:#{null}}")
+  private String operateClientId;
+
+  @Value("${camunda.operate.client.clientSecret:#{null}}")
+  private String operateClientSecret;
+
+  @Value("${camunda.operate.client.enabled:#{true}}")
+  private Boolean operateEnabled;
+
+  // console
+  @Value("${camunda.console.client.clientId:#{null}}")
+  private String consoleClientId;
+
+  @Value("${camunda.console.client.clientSecret:#{null}}")
+  private String consoleClientSecret;
+
+  @Value("${camunda.console.client.enabled:#{true}}")
+  private Boolean consoleEnabled;
+
+  // optimize
+  @Value("${camunda.optimize.client.clientId:#{null}}")
+  private String optimizeClientId;
+
+  @Value("${camunda.optimize.client.clientSecret:#{null}}")
+  private String optimizeClientSecret;
+
+  @Value("${camunda.optimize.client.enabled:#{true}}")
+  private Boolean optimizeEnabled;
+
+  // tasklist
+  @Value("${camunda.tasklist.client.clientId:#{null}}")
+  private String tasklistClientId;
+
+  @Value("${camunda.tasklist.client.clientSecret:#{null}}")
+  private String tasklistClientSecret;
+
+  @Value("${camunda.tasklist.client.enabled:#{true}}")
+  private Boolean tasklistEnabled;
+
+  @Value("${camunda.keycloak.url:#{null}}")
+  private String keycloakUrl;
+
+  @Value("${camunda.keycloak.realm:#{null}}")
+  private String keycloakRealm;
+
+  @Value("${camunda.client.clientId:#{null}}")
+  private String commonClientId;
+
+  @Value("${camunda.client.clientSecret:#{null}}")
+  private String commonClientSecret;
+
+  public Authentication getAuthentication() {
+
+    // check for at-least one existence, then its JWT based authentication
+    if ((zeebeClientId != null ) || (operateClientId != null) || (consoleClientId != null) || (optimizeClientId != null) || (tasklistClientId != null)) {
+      JwtConfig jwtConfig = new JwtConfig();
+
+      if (zeebeEnabled) {
+        if (zeebeClientId != null && zeebeClientSecret != null) {
+          jwtConfig.addProduct(Product.ZEEBE, new Credential(zeebeClientId, zeebeClientSecret));
+        } else if (commonClientId != null && commonClientSecret != null) {
+          jwtConfig.addProduct(Product.ZEEBE, new Credential(commonClientId, commonClientSecret));
+        }
+      }
+
+      if (operateEnabled) {
+        if (operateClientId != null && operateClientSecret != null) {
+          jwtConfig.addProduct(Product.OPERATE, new Credential(operateClientId, operateClientSecret));
+        } else if (commonClientId != null && commonClientSecret != null) {
+          jwtConfig.addProduct(Product.OPERATE, new Credential(commonClientId, commonClientSecret));
+        }
+      }
+
+      if (consoleEnabled) {
+        if (consoleClientId != null && consoleClientSecret != null) {
+          jwtConfig.addProduct(Product.CONSOLE, new Credential(consoleClientId, consoleClientSecret));
+        } else if (commonClientId != null && commonClientSecret != null) {
+          jwtConfig.addProduct(Product.CONSOLE, new Credential(commonClientId, commonClientSecret));
+        }
+      }
+
+      if (optimizeEnabled) {
+        if (optimizeClientId != null && optimizeClientSecret != null) {
+          jwtConfig.addProduct(Product.OPTIMIZE, new Credential(optimizeClientId, optimizeClientSecret));
+        } else if (commonClientId != null && commonClientSecret != null) {
+          jwtConfig.addProduct(Product.OPTIMIZE, new Credential(commonClientId, commonClientSecret));
+        }
+      }
+
+      if (tasklistEnabled) {
+        if (tasklistClientId != null && tasklistClientSecret != null) {
+          jwtConfig.addProduct(Product.TASKLIST, new Credential(tasklistClientId, tasklistClientSecret));
+        } else if (commonClientId != null && commonClientSecret != null) {
+          jwtConfig.addProduct(Product.TASKLIST, new Credential(commonClientId, commonClientSecret));
+        }
+      }
+
+      if (keycloakUrl != null) {
+        return new SelfManagedAuthentication()
+          .jwtConfig(jwtConfig)
+          .keycloakUrl(keycloakUrl)
+          .keycloakRealm(keycloakRealm)
+          .build();
+      } else {
+        return new SaaSAuthentication()
+          .jwtConfig(jwtConfig)
+          .build();
+      }
+    } else {
+      // TODO: SimpleAuthentication and other authentication
+      throw new UnsupportedOperationException("not implemented");
+    }
+  }
+}
--- a/spring-client-common/pom.xml
+++ b/spring-client-common/pom.xml
@@ -12,8 +12,14 @@
 
   <dependencies>
     <dependency>
-      <groupId>io.camunda.spring</groupId>
-      <artifactId>spring-client-annotations</artifactId>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+      <version>5.2.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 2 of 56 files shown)


spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/CommonClientConfigurationProperties.java
```diff
@@ -0,0 +1,146 @@
+package io.camunda.zeebe.spring.client.properties;
+
+import io.camunda.commons.auth.*;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties
```
jonathanlukas, line 7: Could we create configuration properties as classes and not use `@Value` on each property?

This would make the code more readable and the properties classes would properly reflect the property structure of the `application.yaml`.

spring-client-common/src/main/java/io/camunda/commons/auth/SaaSAuthentication.java
```diff
@@ -0,0 +1,104 @@
...
+      throw new RuntimeException("Unable to authenticate");
+    }
+  }
+
+  private void retrieveToken(Product product) {
+    Credential credential = jwtConfig.getMap().get(product);
+    retrieveToken(product, credential);
+  }
+
+  private String getAudience(Product product) {
```
jonathanlukas, line 69: Instead of setting a field, you could just return the audience from inside the case statement

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
```diff
@@ -0,0 +1,115 @@
...
+  private JwtConfig jwtConfig;
+  private Map<Product, String> tokens;
+  private Map<Product, Integer> expirations;
+
+  public SelfManagedAuthentication() {
+    tokens = new HashMap<>();
+    expirations = new HashMap<>();
+  }
+
+  public SelfManagedAuthentication jwtConfig(JwtConfig jwtConfig) {
```
jonathanlukas, line 39: I am not sure how we align on this from DevEx perspective, but I have not seen a builder on the actual class in a Camunda 8 API before..

I would separate the builder from the class and offer a constructor for all args.

This allows for properly setting lifecycle binding of fields (by making them final).

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.stream.Collectors;
+
+import static io.camunda.commons.http.JsonUtils.toResult;
+
+public class SelfManagedAuthentication extends JwtAuthentication {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private String authUrl;
+  private String keycloakRealm = "camunda-platform";
```
jonathanlukas, line 28: For a self-managed setup, there should be other IDPs available soon. An abstraction could be handy here and help to keep the code maintainable

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
```diff
@@ -0,0 +1,115 @@
...
+  private String keycloakUrl;
+  private JwtConfig jwtConfig;
+  private Map<Product, String> tokens;
+  private Map<Product, Integer> expirations;
```
jonathanlukas, line 32: You could make the expiration a LocalDateTime

spring-client-common/pom.xml
```diff
@@ -12,8 +12,14 @@
...
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+      <version>5.2.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.13.1</version>
     </dependency>
   </dependencies>
```
sbuettner, line 24: Looking at the dependencies we could offer our clients independently from spring:
`camunda-sdk-java` and re-use them in a `camunda-sdk-spring` offering. 
This would unblock users from using the clients with other frameworks like quarkus, ktor etc.

spring-client-common/src/main/java/io/camunda/commons/auth/SaaSAuthentication.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private void retrieveToken(Product product, Credential credential) {
+    try {
+      HttpPost httpPost = new HttpPost(authUrl);
+      httpPost.addHeader("Content-Type", "application/json");
+      TokenRequest tokenRequest = new TokenRequest(getAudience(product), credential.clientId, credential.clientSecret);
+
+      httpPost.setEntity(new StringEntity(toJson(tokenRequest)));
+      CloseableHttpClient client = HttpClients.createDefault();
```
sbuettner, line 53: We could re-use an existing client here instead of creating a new one for every token request.

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
```diff
@@ -0,0 +1,115 @@
...
+        })
+        .collect(Collectors.joining("&"));
+
+      httpPost.setEntity(new StringEntity(form));
+      CloseableHttpClient client = HttpClients.createDefault();
+      CloseableHttpResponse response = client.execute(httpPost);
+      TokenResponse tokenResponse = toResult(EntityUtils.toString(response.getEntity()), TokenResponse.class);
+      // TODO: verify JWT has the desired permission vs what the user requested
+      tokens.put(product, tokenResponse.getAccessToken());
+      expirations.put(product, (int) (System.currentTimeMillis()/1000 + tokenResponse.getExpiresIn()));
```
sbuettner, line 90: Why do we convert it to seconds when we have to convert it again during refresh? Maybe just store it in `ms`.

spring-client-common/src/main/java/io/camunda/commons/http/JsonUtils.java
```diff
@@ -0,0 +1,60 @@
...
+    private static Map<Class<?>, JavaType> searchResultTypeMap = new HashMap<>();
+
+    private static ObjectMapper getObjectMapper() {
+        if (mapper == null) {
+            mapper = new ObjectMapper();
+            mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+            mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        }
+        return mapper;
+    }
```
sbuettner, line 30: I would recommend to pass a simplified object mapper interface like the `JsonMapper` in the zeebe java client: https://github.com/camunda/zeebe/blob/main/clients/java/src/main/java/io/camunda/zeebe/client/api/JsonMapper.java and default to the Jackson `ObjectMapper`.

spring-client-common/src/main/java/io/camunda/commons/http/HttpClient.java
```diff
@@ -17,7 +18,8 @@ public interface HttpClient {
 
   <T> T get(Class<T> clazz, String id);
 
-  void get(String resource);
+  <T> String getXml(Class<T> clazz, Long key);
+
+  <T, V, W, U> T post(Class<T> clazz, Class<V> vj, TypeToken<W> typeToken, U body);
```
jonathanlukas, line 23: Could the parameter be a bit more explicit please? I do not understand what `vj` should be and what the typeToken is good for.

1nb0und: My bad. I will make it more explicit. Was committing things yesterday and didn't realize i left some test names out there.

spring-client-common/src/main/java/io/camunda/commons/auth/SaaSAuthentication.java
1nb0und: Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/e7618e57f74828076725ca17a2404687f0b0e506

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
1nb0und: Separated the builder from the class in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/603bb6a40b1d5295d7d805c0648569b90972389c

However I don't get where do you suggest to put the all args constructor. I guess that goes in the class itself? So we would offer users two ways to use a class: either thru the builder or all args. 

I don't quite get the `property setting lifecylcle bindings of fields`. What advantage or use would that give us? Any examples maybe from Zeebe?

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
1nb0und: Not still familiar with IDP and Identity world. I would meet with Identity team soon and will check more about upcoming IDPs.

Adding a TODO for now in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/0922e0fe0feb5c4c0f8a6339d7d99b79ecfeefc1

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
1nb0und: Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/0b537aa7a8ff6903bea1a85a915e2f184495f95d

spring-client-common/pom.xml
1nb0und: I like the idea, I'll check how to best structure the project modules to offer the `camunda-sdk-java`. Although the entire repo will be evolving first to Spring first SDK. 

I can probably see what can be done so that the main repo will be about Spring SDK, and then we can have a tiny or special module that can cater to non-spring apps but not explicitly (kinda similar on how we embed `Operate` in Spring-Zeebe and can offer it as a bean but we are not openly advertising it).

spring-client-common/src/main/java/io/camunda/commons/auth/SelfManagedAuthentication.java
1nb0und: Switched to `LocalDateTime` and got rid of seconds / milliseconds in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/0b537aa7a8ff6903bea1a85a915e2f184495f95d

spring-client-common/src/main/java/io/camunda/commons/http/JsonUtils.java
1nb0und: Added in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/092480595ac84d59911023f2e79a6545153a36da

spring-client-common/src/main/java/io/camunda/commons/http/HttpClient.java
1nb0und: Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/45441794dbc1d7a37c1d24fb2cf1ed00aaa41e11

spring-client-common/pom.xml
sbuettner: Yes, that would be great. I can imagine to just have a json+http dependency in a java sdk for example.

spring-client-common/src/main/java/io/camunda/commons/auth/SaaSAuthentication.java
1nb0und: Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/4036ec7c35e84168e582d91e2ff44edec6d50240

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/CommonClientConfigurationProperties.java
1nb0und: Got rid of `@Value`s and refactored as its own configuration properties in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/2a13aa5a78ea11c117eae5e356615b180113bc0a

spring-client-common/pom.xml
1nb0und: Ok I caved in @sbuettner  😁  Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/510/commits/f5e117b341f1297526050f79f8cbbf25ce19e66f. 

And I think we are on the right track.

So essentially it will now look like this

Screenshot 2023-11-22 at 4 06 05 PM

We now have a plain Java module to support clients other than Spring. But first priority will be given for Spring related tasks. 

I noticed that there is no sense to maintain the `spring-client-operate` module so removed it for now (since the `c8-java-sdk/java-client-operate`) and the `starter-camunda` handles the auto-configuration. If there is a need for custom annotations for `Operate` which I don't foresee at the moment, I can bring it back.

berndruecker: +100! And just to double down: We should avoid "c8" as a name anyway - it is Camunda 8 or simply Camunda :-)

Issue #511, opened 2023-11-13
Work with Spring Boot 3.2


BalikhinLB:
The release of Spring Boot 3.2.0 is scheduled for November 23rd. In it, the class [LocalVariableTableParameterNameDiscoverer](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/LocalVariableTableParameterNameDiscoverer.html) will be removed. To ensure functionality, it is necessary to update the MethodInfo class, replacing LocalVariableTableParameterNameDiscoverer with [StandardReflectionParameterNameDiscoverer](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/StandardReflectionParameterNameDiscoverer.html). Additionally, to ensure successful operation, it is necessary to update the pom.xml file:
`
<plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-compiler-plugin</artifactId>
          <version>3.11.0</version>
          <configuration>
          <source>${java.version}</source>
          <target>${java.version}</target>
            <compilerArgs>
              <arg>-parameters</arg>
            </compilerArgs>
          </configuration>
</plugin>
`
The task is not a big one, I could have done it myself. However, it seems that some permissions are needed to push the branch, I am currently getting a 403 error.

berndruecker:
Can you create a Pull Request instead @BalikhinLB ?

BalikhinLB:
I feel very embarrassed, I haven't contributed to open source yet. To make a pull request here, do I need to execute a fork? I don't have the option to pull the branch directly into the current repository?

berndruecker:
No worries - the first time is always the most exciting one :-)
Yes - you need to fork the repo and do the changes in your fork (there you have all permissions). Then you have the option to create a PR from there. Typically you create a branch for each PR - but I don't think GitHub forces you to.
See also https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

BalikhinLB:
Ok, thx. I did it: [PR](https://github.com/camunda-community-hub/spring-zeebe/pull/512)

PR #512, opened 2023-11-13
fix parameterNameDiscoverer for work with spring boot 3.2.0

BalikhinLB:
pull request for [issues-511](https://github.com/camunda-community-hub/spring-zeebe/issues/511)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -170,13 +170,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.11.0</version>
+          <!-- available from Java 9, but we went back to jdk8 for building
           <configuration>
-            <source>${java.version}</source>
-            <target>${java.version}</target>
-            <compilerArgs>
-              <arg>-parameters</arg>
-            </compilerArgs>
+            <release>${java.version}</release>
           </configuration>
+      	  -->
         </plugin>
         <plugin>
             <groupId>org.apache.maven.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -170,11 +170,13 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.11.0</version>
-          <!-- available from Java 9, but we went back to jdk8 for building
           <configuration>
-            <release>${java.version}</release>
+            <source>${java.version}</source>
+            <target>${java.version}</target>
+            <compilerArgs>
+              <arg>-parameters</arg>
+            </compilerArgs>
           </configuration>
-      	  -->
         </plugin>
         <plugin>
             <groupId>org.apache.maven.plugins</groupId>
--- a/test/embedded/pom.xml
+++ b/test/embedded/pom.xml
@@ -41,7 +41,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.11.0</version>
         <configuration>
           <source>${java.version}</source>
           <target>${java.version}</target>
--- a/test/embedded/pom.xml
+++ b/test/embedded/pom.xml
@@ -36,4 +36,21 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.11.0</version>
+        <configuration>
+          <source>${java.version}</source>
+          <target>${java.version}</target>
+          <compilerArgs>
+            <arg>-parameters</arg>
+          </compilerArgs>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
--- a/spring-client-zeebe/pom.xml
+++ b/spring-client-zeebe/pom.xml
@@ -54,7 +54,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.11.0</version>
         <configuration>
           <source>${java.version}</source>
           <target>${java.version}</target>
--- a/spring-client-zeebe/pom.xml
+++ b/spring-client-zeebe/pom.xml
@@ -49,4 +49,20 @@
     </dependency>
 
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.11.0</version>
+        <configuration>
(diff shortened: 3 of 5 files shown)


pom.xml
```diff
@@ -170,11 +170,13 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.11.0</version>
-          <!-- available from Java 9, but we went back to jdk8 for building
```
berndruecker, line 173: @BalikhinLB Is this change related to the relevant one or could that be reverted?

berndruecker (changes requested):
LGTM beside the changes in the pom.xml that should probably not be mixed into the PR

pom.xml
BalikhinLB: This change is required for the application to work.
inside parameterNameDiscoverer.getParameterNames(method):
```
if (!param.isNamePresent()) {
                return null;
}
```
  and this change need for param.isNamePresent() = true

BalikhinLB: You can see it [here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/StandardReflectionParameterNameDiscoverer.html)

pom.xml
```diff
@@ -170,11 +170,13 @@
...
           <configuration>
-            <release>${java.version}</release>
+            <source>${java.version}</source>
+            <target>${java.version}</target>
+            <compilerArgs>
+              <arg>-parameters</arg>
```
1nb0und, line 177: Is this working at your end? I tried your changes but it seems the `parameterNames` always end up being `null` despite the switch from `LocalVariableTableParameterNameDiscoverer` to `StandardReflectionParameterNameDiscoverer`. Wondering if there is anything else that needs to be refactored.

pom.xml
1nb0und: @berndruecker  I think this change is required. It seems the new `StandardReflectionParameterNameDiscoverer ` requires some additional compiler help . This [article](https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html) helps explains it.

My only concern would be, it seems due to this new required compiler flag, we would require our new and existing customers to add this new flag on their build. 

Our customers would use their own projects and create their own Zeebe Workers and they annotate it with `@Variable` for the method parameters. For all of this to work, the compiler flag needs to be enabled on their respective projects. I'll verify that, and see if thats the case. This would be a breaking change to them and docs updates might be needed.

pom.xml
1nb0und: I stand corrected. I forgot to recompile my classes with the new flag added. It now works!

pom.xml
1nb0und: Yes I can confirm that if you are using Zeebe workers and utilizing parameter names such as `@Variable` then customers would be required to compile their code with `parameters`.  We gotta need to think more on this one regarding the user-experience. 

@BalikhinLB I think we can remove the configuration on the parent pom.xml, and just put it under the `spring-boot-starter-camunda-example`

image

It looks like this is only required on code that uses the custom Zeebe worker annotations. Other modules doesn't need this.

BalikhinLB: The ReadZeebeWorkerValueTest test is failing because it checks the MethodInfo from ClassInfoTest.WithZeebeWorker.class, and it is precisely here that parameters are needed. The JobHandlerTest, SmokeTest are also failing.
To avoid test failures, parameters need to be added in spring-client-zeebe, test/embedded, and test/testcontainer. If you think this is okay - I will do so. Honestly, I don't really understand why java.version can be 1.8 when spring now practically (except for the most ancient versions) only works on Java 11+ versions.

1nb0und: @BalikhinLB Good catch. Yeah please apply the parameters to above affected modules.

BalikhinLB: @1nb0und ok, i applied this

1nb0und: I think this change is good to go in. Also updated readme in this PR https://github.com/camunda-community-hub/spring-zeebe/pull/514 so users are aware when using new Spring-Zeebe versions higher than 8.3.1

@berndruecker thoughts on this?

berndruecker: Bummer that this is required to continue having the same functionality, as this is really a breaking change from user perspective. But I agree that this is probably OK to do now - also good to add it to the readme (I commented in #514 - I think we should specifically say that this is only required for parameter annotations to work).

I wonder if this is just a Spring related thing or if this is generally required to process parameter annotiations - but don't have time to dive into this and leave it to your judgement @1nb0und . 

In short: yeah - ok for me

test/embedded/pom.xml
```diff
@@ -36,4 +36,21 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.11.0</version>
```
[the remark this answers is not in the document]
BalikhinLB: Guys, seriously, of course, this might work and will work and I even fixed it, but the pull request is not about that at all.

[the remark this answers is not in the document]
1nb0und: @BalikhinLB the extra versions set was part of the PR and code quality is important too in the community. Please take this as a positive feedback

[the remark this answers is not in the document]
BalikhinLB: @1nb0und I apologize. I lost focus at the end of the day. I indeed wrote the code, which means I messed up. Thanks for the note, I corrected my mistake, not the "as it was".

dvag-martin-anlauf:
Hello, when we update our project from spring boot 3.1.6 to 3.2.0 we get problems during bean creation because two beans of the same type but with different names can't be distinguished any more. We get a messege like 'Unsatisfied dependency expressed through method 'xy' parameter 1: No qualifying bean of type 'ba.Ba' available: expected single matching bean but found 2: defRestTemplate,restTemplate'.
It happens, because the LocalVariableTableParameterNameDiscoverer is no more available to the PrioritizedParameterNameDiscoverer, so it should be the effect discussed here.
Do we have a possibility to correct the behavior ourself using spring boot 3.20? I.e. to enable again to distinguish beens according to their names? Or do we have to skip spring boot 3.2.0 and wait for the next version 3.2.1 where it is fixed? I did not understand that point.
Any help appreciated, thank you.

dvag-martin-anlauf:
Thank you and sorry for polluting your PR.  Basically, our problem was related to spring-core, not to camunda. Nevertheless the discussion was helpful and we could solve the problem. 
We have to activate the "-parameters" option for compilation. With that option set the parameter names become available again to resolve ambiguities in case of several Beans of the same type.
We just had to add a plugin configuration in maven:
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <configuration>
                    <parameters>true</parameters>
                    <compilerArgs>
                        <arg>-parameters</arg>
                    </compilerArgs>
                </configuration>
            </plugin>

PR #514, opened 2023-11-15
Adding instruction for new compiler flag

1nb0und:
Updating the readme to reflect usage of new compiler flag for new versions of Spring-Zeebe related to this PR https://github.com/camunda-community-hub/spring-zeebe/pull/512

This is the first time i'm aware of the `-parameters` flag but it seems its common based on a[ quick github search](https://github.com/search?q=%3Carg%3E-parameters%3C%2Farg%3E+&type=code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,39 @@ Although Spring Zeebe has a transitive dependency to the [Zeebe Java Client](htt
 </dependency>
 ```
 
+The following compiler flag `-parameters` is required for Spring-Zeebe versions higher than 8.3.1:
+
+If using Maven in your `pom.xml`
+```xml
+<build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,38 @@ Although Spring Zeebe has a transitive dependency to the [Zeebe Java Client](htt
 </dependency>
 ```
 
+The following compiler flag `-parameters` is required for Spring-Zeebe versions higher than 8.3.1:
```
berndruecker, line 73: but only if you want to use @VAriables - right? Then I would be specific here - then users can decide to not add that flag and just not use Variables

1nb0und: Yup! Added explicit note [here](https://github.com/camunda-community-hub/spring-zeebe/pull/514/commits/751b318461526dd060a769764a32931f833c2732)

christinaausley (approved):

PR #554, opened 2023-12-13
Support old and new operate properties schema

1nb0und:
We need to support both old and new operate properties schema

Old:

`camunda.operate.client.*`

New:

`operate.client.*`

Until then, we will deprecate and remove the older convention in favor of the newer one

# Changes
- support both conventions
- update readme
- remove explicitly enable = true (derive it based on other properties set)


sbuettner (approved):
Thanks for taking care @1nb0und.

I couldnt find any test so I assume its covered by existing tests.

PR #562, opened 2023-12-15
Revert to Java 17

1nb0und:
This PR will
- specify Java 21 only for embedded tests
- root project is at Java 17 (and so the rest of the projects will be at 17)
- Maven build and release robots will use Java 21 as one of our modules (embedded) needs 21 (even if build process is at higher JDK, we force the maven modules to use specified version (17) and should be backwards compatible)


cbertoldi:
I know that this merge has been reverted in the 8.4.0, but I just tried building my application with the 8.4.1 and got the following error:

`cannot access io.camunda.zeebe.spring.test.ZeebeSpringTest
[ERROR]   bad class file: ~/.m2/repository/io/camunda/spring/spring-boot-starter-camunda-test/8.4.1/spring-boot-starter-camunda-test-8.4.1.jar(/io/camunda/zeebe/spring/test/ZeebeSpringTest.class)
[ERROR]     class file has wrong version 65.0, should be 61.0
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.`

If I switch back to 8.3.5 the error does not occur.

cbertoldi:
I was too quick asking, I found this issue https://github.com/camunda-community-hub/spring-zeebe/issues/623 and it has been merged only yesterday, so I guess I will have to wait for the 8.4.2, correct?

PR #579, opened 2023-12-22
Create Operate client README.md with a few examples

chDame:


camunda-sdk-java/java-client-operate/README.md
```diff
@@ -0,0 +1,69 @@
...
+  JwtConfig jwtConfig = new JwtConfig();
+  jwtConfig.addProduct(Product.OPERATE, new JwtCredential(clientId, clientSecret));
+  auth = SelfManagedAuthentication.builder().jwtConfig(jwtConfig).keycloakUrl(keycloakUrl).build();
+
+  client = CamundaOperateClient.builder()
+              .operateUrl(operateUrl)
+              .authentication(auth)
+              .setup()
+              .build();
+```
```
1nb0und, line 43: I just recently updated and fixed the SimpleAuthentication. Do you want to include it as well? Or I can include it after?

1nb0und (approved):
LGTM 👍

camunda-sdk-java/java-client-operate/README.md
chDame: Yes, I'll let you do. I'm wondering if we should document it in the java-common and just link it from the client to not repeat ourselves when other clients will be integrated

PR #587, opened 2024-01-05
chore(build): Add missing compiler arg

sbuettner:
Add compiler flag to avoid https://github.com/camunda-community-hub/spring-zeebe/pull/514 when running tests for `spring-boot-starter-camunda`


1nb0und (approved):
Good catch!

PR #588, opened 2024-01-05
Add token reset and throw exception instead of swallowing

1nb0und:
- Add token reset for authentication failures
- Throw exceptions instead of swallowing

TODOs:
- Refactor the duplicate code introduced
- Handle automatic retry of existing request if authentication has expired
- Make code unit testable to add unit tests and mock authentication process

### DIFF
--- a/camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SimpleAuthentication.java
+++ b/camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SimpleAuthentication.java
@@ -44,7 +44,6 @@ public class SimpleAuthentication implements Authentication {
   @Override
   public Authentication build() {
     authUrl = simpleUrl+"/api/login";
-    //simpleConfig.getMap().forEach(this::retrieveToken);
     return this;
   }
 
@@ -88,5 +87,11 @@ public class SimpleAuthentication implements Authentication {
     }
 
     return new AbstractMap.SimpleEntry<>("Cookie", token);
+  }
+
+  @Override
+  public void resetToken(Product product) {
+    SimpleCredential simpleCredential = simpleConfig.getProduct(product);
+    retrieveToken(product, simpleCredential);
   }
 }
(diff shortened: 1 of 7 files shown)


camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SimpleAuthentication.java
```diff
@@ -89,4 +88,10 @@ public Map.Entry<String, String> getTokenHeader(Product product) {
 
     return new AbstractMap.SimpleEntry<>("Cookie", token);
   }
+
+  @Override
+  public void resetToken(Product product) {
+    SimpleCredential simpleCredential = simpleConfig.getProduct(product);
+    retrieveToken(product, simpleCredential);
```
sbuettner, line 95: Why not just remove the token from `tokens` to make sure an invalid token is removed?

camunda-sdk-java/java-common/src/main/java/io/camunda/common/http/DefaultHttpClient.java
```diff
@@ -116,12 +119,13 @@ public <T> String getXml(Class<T> selector, Long key) {
     String url = host + basePath + retrievePath(selector) + "/" + key + "/xml";
     HttpGet httpGet = new HttpGet(url);
     httpGet.addHeader(retrieveToken(selector));
-    String xml = null;
+    String xml;
     try {
       CloseableHttpResponse response = httpClient.execute(httpGet);
-      xml = new String(Java8Utils.readAllBytes(response.getEntity().getContent()), StandardCharsets.UTF_8);
+      xml = (String) parseAndRetry(response, selector);
```
sbuettner, line 125: As discussed this will break as it will try to parse the XML as JSON.

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SimpleAuthentication.java
1nb0und: Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/588/commits/137ff5b93dc9b589063987d6edf39b9418cf9841

camunda-sdk-java/java-common/src/main/java/io/camunda/common/http/DefaultHttpClient.java
1nb0und: Addressed in https://github.com/camunda-community-hub/spring-zeebe/pull/588/commits/bc36a55b3e21845ca2d13816e30eb145cd98f9ee

PR #596, opened 2024-01-10
fix(http): Close the http client resources in the SDK authentication

sbuettner:


chillleader (approved):

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SaaSAuthentication.java
```diff
@@ -53,23 +53,27 @@ public void resetToken(Product product) {
...
+      } catch (Exception e) {
       LOG.error("Authenticating for " + product + " failed due to " + e);
       throw new RuntimeException("Unable to authenticate", e);
     }
     return tokens.get(product);
   }
 
+  private HttpPost buildRequest(JwtCredential jwtCredential) {
+    HttpPost httpPost = new HttpPost(jwtCredential.getAuthUrl());
+    httpPost.addHeader("Content-Type", "application/json");
```
igpetrov, line 71: nitpick: pretty sure there are constants for the content type, maybe `HttpHeaders.CONTENT_TYPE` and  `MediaType.JSON_UTF_8.toString()`.

sbuettner: Yes, I wanted to keep the changes minimal but will tag @1nb0und so he could incorporate your suggestions.

PR #600, opened 2024-01-10
fix(http): remove cache http client

1nb0und:
Removes the cache http client that can be exhausted and never replaced. Now we will create a new instance for each request.


chillleader (approved):

igpetrov (approved):
LGTM, thank you

PR #604, opened 2024-01-19
added console client

jonathanlukas:
This PR will add a generated console client, wrap it with a handsome api and also add it to the spring boot starter.

It is based on https://github.com/camunda-community-hub/camunda-console-client-java which will be archived after this PR is merged.


camunda-sdk-java/java-client-console/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+      <plugin>
+        <groupId>org.openapitools</groupId>
+        <artifactId>openapi-generator-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>generate</goal>
+            </goals>
+            <configuration>
+              <inputSpec>https://console.cloud.camunda.io/customer-api/openapi/swagger.json</inputSpec>
```
1nb0und, line 79: This is some cool stuff. I never used a generator before. I need to see if shall we go the generator route, or write out the API models ourselves. Wondering what are the advantages and disadvantages.

The route I've been going is to manually add the classes and use sensible names (since TaskList, Operate and others from what I've seen are using different convention). I understand that going thru the generated route reduces time to market, and maintenance I assume would be easier since whenever they update their version we can just generate again and modify the wrapper classes calling it.

camunda-sdk-java/java-client-console/src/main/java/io/camunda/console/DefaultApiFactory.java
```diff
@@ -0,0 +1,42 @@
...
+      HttpRequestInterceptor interceptor =
+          new AuthenticationInterceptor(Product.CONSOLE, authentication);
+      CloseableHttpClient httpClient =
+          HttpClients.custom().addRequestInterceptorFirst(interceptor).build();
+      ApiClient apiClient = new ApiClient(httpClient);
+      apiClient.setBasePath(consoleUrl);
+      api = new DefaultApi(apiClient);
+    }
+  }
+}
```
1nb0und, line 42: I'd like to reuse the existing `java-sdk-common.http` package for this rather than creating another HttpClient

camunda-sdk-java/java-client-console/pom.xml
jonathanlukas: I think we can go the generator route for all APIs that offer some kind of spec. 

I see that this kind of refactoring would probably cost some effort, but also reduce the amount of code to be maintained.

1nb0und (commented):
This is looking awesome. I think I want to improve a couple of things by

1. see if we still want to use the generator approach (the other modules are hand-crafted models). We can go the route of this being generator, and the remaining are hand-crafted and see how will versioning and maintenance will look like whenever there is a change in the API
2. I'd like to reuse the existing `http` module. Theres some refactoring involved.

I'm more than happy to do those changes, but wanted to get your opinion. Overall this speeds things up and this is greatly valued!

camunda-sdk-java/java-client-console/src/main/java/io/camunda/console/DefaultApiFactory.java
jonathanlukas: The HttpClient you created is not compatible with the generated client as this one relies on the selected http client framework (apache in this case).

This is why I added the AuthenticationInterceptor.

jonathanlukas: I understand that your own http client helps us being independent of any http client framework, however the amount of code required to bind our implementation against a framework is REALLY limited (like 1 class with 10 loc)

jonathanlukas:
@1nb0und (and maybe @berndruecker ):

Maybe my 2 cents for generated client vs. implemented client:

As an API is implemented API-first (define open api spec, generate server stub and implement it), there is a revisioned resource openapi.json available to work with. Here, a generated client makes sense.

If however, the API is implemented impl-first (write server stub, generate open api spec at runtime), there is no real resource available to implement against. Here, we could rely on an implemented client.

wdyt?

PR #606, opened 2024-01-20
Identity SDK Integration

1nb0und:
Integrates Identity SDK for Self Managed authentication


camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
```diff
@@ -81,65 +63,22 @@ public void resetToken(Product product) {
...
+      String audience = jwtConfig.getProduct(product).getAudience();
+      Tokens identityTokens = identity.authentication().requestToken(audience);
+      try {
+        identity.authentication().verifyToken(identityTokens.getAccessToken());
+      } catch (TokenExpiredException exception) {
+        identityTokens = identity.authentication().renewToken(identityTokens.getRefreshToken());
+      }
+      tokens.put(product, identityTokens.getAccessToken());
+      token = tokens.get(product);
     }
```
sbuettner, line 82: I would recommend to extract this into a separate method.

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
```diff
@@ -81,65 +63,22 @@ public void resetToken(Product product) {
...
-      JwtCredential jwtCredential = jwtConfig.getProduct(product);
-      token = retrieveToken(product, jwtCredential);
+      Identity identity = identityConfig.get(product).getIdentity();
+      String audience = jwtConfig.getProduct(product).getAudience();
+      Tokens identityTokens = identity.authentication().requestToken(audience);
+      try {
+        identity.authentication().verifyToken(identityTokens.getAccessToken());
+      } catch (TokenExpiredException exception) {
+        identityTokens = identity.authentication().renewToken(identityTokens.getRefreshToken());
+      }
```
sbuettner, line 79: Do we expect that a token fetched just now is already expired?

1nb0und: Thats a remote possibility but these verify logic here can be removed if Identity SDK has an ability to remove the token in their cache. Raised a [question](https://github.com/camunda-cloud/identity/issues/2496)/feature request on their board

Basically, whenever we received a 401 or expired token from some endpoints, we remove the token in memory in Spring SDK. However since we now rely on Identity SDK, we would also need to have the ability to evict token in Identity Cache too.

Right now, I'm attempting to remove the token in Spring SDK memory, however whenever we get a new token with Identity that was already expired, their logic would still retrieve the expired in cache. The current workaround would be get a new token thru referesh token, but as far as I know, the expired token always sits in their cache.

sbuettner: I was just wondering as this code is only called after fetching a new token.

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
1nb0und: [Applied](https://github.com/camunda-community-hub/spring-zeebe/pull/606/commits/ffa8cf4f11c5fd958d733f903ed9aa2f3c006512)

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
```diff
@@ -69,17 +69,21 @@ public Map.Entry<String, String> getTokenHeader(Product product) {
...
+    Identity identity = identityConfig.get(product).getIdentity();
+    String audience = jwtConfig.getProduct(product).getAudience();
+    Tokens identityTokens = identity.authentication().requestToken(audience);
+    try {
+      identity.authentication().verifyToken(identityTokens.getAccessToken());
+    } catch (TokenExpiredException exception) {
+      identityTokens = identity.authentication().renewToken(identityTokens.getRefreshToken());
+    }
+    tokens.put(product, identityTokens.getAccessToken());
+    return tokens.get(product);
```
sbuettner, line 87: I would recommend to only handle the token creation here and put this part of the coder in the caller that is responsible for managing the tokens.

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
```diff
@@ -69,17 +69,21 @@ public Map.Entry<String, String> getTokenHeader(Product product) {
...
+
+  private String getIdentityToken(Product product) {
+    Identity identity = identityConfig.get(product).getIdentity();
+    String audience = jwtConfig.getProduct(product).getAudience();
+    Tokens identityTokens = identity.authentication().requestToken(audience);
+    try {
+      identity.authentication().verifyToken(identityTokens.getAccessToken());
+    } catch (TokenExpiredException exception) {
+      identityTokens = identity.authentication().renewToken(identityTokens.getRefreshToken());
+    }
```
sbuettner, line 85: I still dont get why we renew token we just created here?  Or does the `identity` cache it internally?

I would recommend to take a look at the Identity SDK internals: https://github.com/camunda-cloud/identity/blob/db2dae3da92bbbfc34e646297e4bdd6ba04f07ec/sdk-java/src/main/java/io/camunda/identity/sdk/authentication/AbstractAuthentication.java#L80

1nb0und: So the issue this is trying to solve is because of the following problem:

Identity code internals[ doesn't provide a way ](https://github.com/camunda-cloud/identity/issues/2496#issuecomment-1904273555)for its users to remove expired cache token internally when you are using the Identity SDK. I requested Identity team to expose a method that will either delete the token or refresh it. 

If we don't have the verify / renew token code here (L81-L85), what would happen is we would keep getting expired tokens since `identityTokens.getAccessToken()` always pull it from their cache and its not refreshing it automatically.

> I still dont get why we renew token we just created here?

We renew it since Identity internals doesn't renew their tokens when first created. So even if the method here says `getIdentityToken(Product product)`, it always pull from cache. It is our responsibility to check if it was expired and if its then refresh it.

Once Identity provided a method to get a fresh and update their internal cache, then the entire `try-catch` block can be remove, and thus avoids us checking if token is expired


Does this make sense?

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
1nb0und: I agree, the function is doing multiple things at once. Refactored [here](https://github.com/camunda-community-hub/spring-zeebe/pull/606/commits/305d96c91b8840fc082ccca94f3550bb6ad1b912)

camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/SelfManagedAuthentication.java
sbuettner: Ok, thanks for the explanation. I thought we might rather need to do it when a client receives an error from the server. Not sure why the SDK would return an invalid token from the cache.

Could you create a ticket in the Identity SDK for this issue: https://github.com/camunda-cloud/identity/issues so the SDK can provide this functionality?
cc @Ben-Sheppard

1nb0und: Raised the [issue](https://github.com/camunda-cloud/identity/issues/2496) on Identity

1nb0und: I would also need to update some of the code here since I got to know that M2M calls will never have refreshTokens so this line has to go away

`identityTokens = identity.authentication().renewToken(identityTokens.getRefreshToken())`

Will refactor and do some testing for expiration scenarios for this

PR #613, opened 2024-01-25
feat(config): Re-introduce toString() methods for configuration

jonathanlukas:
Before the refactoring of the client configuration, the whole config was printed on startup.

This PR re-introduces this to verify whether all configuration has been applied correctly.

### DIFF
--- a/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfiguration.java
+++ b/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfiguration.java
@@ -199,4 +199,9 @@ public class ZeebeClientConfiguration implements io.camunda.zeebe.client.ZeebeCl
   public boolean useDefaultRetryPolicy() {
     return properties.useDefaultRetryPolicy();
   }
+
+  @Override
+  public String toString() {
+    return "ZeebeClientConfiguration{" + "properties=" + properties + ", commonConfigurationProperties=" + commonConfigurationProperties + ", authentication=" + authentication + ", jsonMapper=" + jsonMapper + ", interceptors=" + interceptors + ", zeebeClientExecutorService=" + zeebeClientExecutorService + '}';
+  }
 }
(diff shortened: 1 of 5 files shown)


spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfiguration.java
```diff
@@ -200,4 +200,8 @@ public boolean useDefaultRetryPolicy() {
     return properties.useDefaultRetryPolicy();
   }
 
+  @Override
+  public String toString() {
+    return "ZeebeClientConfiguration{" + "properties=" + properties + ", commonConfigurationProperties=" + commonConfigurationProperties + ", authentication=" + authentication + ", jsonMapper=" + jsonMapper + ", interceptors=" + interceptors + ", zeebeClientExecutorService=" + zeebeClientExecutorService + '}';
```
sbuettner, line 205: Could this potentially leak credentials?

jonathanlukas: True. Set all to hidden as far as I could identify them, see latest commit

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/properties/ZeebeClientConfigurationProperties.java
```diff
@@ -305,20 +332,42 @@ public boolean equals(Object o) {
...
+  public static class Cloud {
 
     @Override
     public String toString() {
-      return "Broker{"
-          + "gatewayAddress='"
-          + gatewayAddress
+      return "Cloud{"
+          + "clusterId='"
+          + "***"
```
sbuettner, line 343: Cluster ID could be useful wdyt?

sbuettner, line 343:
```suggestion
          + "clusterId='"
          + clusterId
```

sbuettner (approved):

PR #614, opened 2024-01-26
Support multi-tenancy per worker

### LINKED ISSUE #501: Add multi-tenancy support
We should add recently landed multi-tenancy support of the Java Client in Spring Zeebe.

Currently I am mostly thinking of fetching jobs for one tenant only: https://javadoc.io/doc/io.camunda/zeebe-client-java/latest/io/camunda/zeebe/client/api/command/CommandWithOneOrMoreTenantsStep.html:

```java
client.newJobWorker()
    .jobType("foo")
    .handler(handler)
    .tenant("xy")
    .open();
```

But I might miss something, haven't dived into that myself yet.

jonathanlukas:
As of now, mulit-tenancy is only supported for the whole client.

This PR enables the usage of tenants on worker level.

This PR also refactors the annotation processor so that default job worker name and type are not injected from the spring environment. Instead, this was moved to the PropertyBasedZeebeWorkerValueCustomizer as here, the properties are available directly.

Hint: This PR also introduces the usage of a formatter in the maven build process. All formatter changes are in separate commits.

closes #501

### DIFF
--- a/.github/workflows/mvn-build.yml
+++ b/.github/workflows/mvn-build.yml
@@ -20,4 +20,4 @@ jobs:
           java-version: '21'
           distribution: 'adopt'
       - name: Build with Maven
-        run: mvn --batch-mode --update-snapshots package
+        run: mvn verify -PcheckFormat -B
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ If you have additional information that can help us understand the behaviour of 
 
 If no issue already exists for the bug, please create a new one.
 
-When creating a bug report, please fill out the template. What really helps us are the steps for us to be able reproduce the problem in front of us. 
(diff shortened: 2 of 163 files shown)


sbuettner (commented):
Hard to judge as this part is not covered by tests.
@jonathanlukas Did you test it with an existing application

sbuettner:
Great that you integrated a formatter!

jonathanlukas:
@sbuettner there is now a test in place (see last commit) which asserts that the ZeebeWorkerValue picks up the tenant ids.

For 2e2 coverage, there would have to be a way to look at the actual job worker configuration or intercept a activateJobs command by the client...

sbuettner (approved):

sbuettner:
Thanks @jonathanlukas

PR #656, opened 2024-02-21
improved parameter injection

### LINKED ISSUE #653: Rework variable mapping
Currently, all variable mapping is calculated at runtime. This means that reflections, etc. is done for each invocation.
Also, variable can only be performed for process engine data.

The solution would address these 2 issues:

Providing an interface to map variables from various data sources (including the job variables) to the job worker. All mappings are calculated upfront and prepared mapping functions are invoked on each job worker invocation

Leave out the part with refactoring and focus on data mapping extensibility.

**Additional context**
no

jonathanlukas:
This PR moves the calculation of which parameter of a @JobWorker needs to be injected how from execution to instantiation.

This new mechanism is called parameter resolving.

Also, this PR adds extensibility to this parameter resolving by injecting a custom strategy to the job worker manager instead of the default one.

closes #653


chillleader (approved):
LGTM 🙌

PR #690, opened 2024-03-19
beackport 8.5: Feature/reorganize properties (#662)

jonathanlukas:
* deprecate old properties

* refactoring usage of properties

* wip

* wip

* wip

* wip

* legacy config still working

* removed unused configuration

* added backend issuer url to default saas props

* adjusted readme

* fixed broken test

* functionality of new property strategy proved

* moved from spring profiles to properties post processor

* moved tests as well

* worker overrides are applied over global overrides

* typo

* table format

* add a general comment for first orientation

* removed property

* typo

* updated docs about variable fetching

* typo

* update link to reflect new properties class

* typo

* deprecated classes

* updated executorserviceconfig to reflect new properties

* added a test to assert all deprecations are applied correctly

* applied deprecation notices to all legacy properties

* adjusted a reason comment

(cherry picked from commit ef375370c2c326055b2e0d1078dd015ca8efe90b)


chillleader:
@jonathanlukas Should we perhaps include it in 8.6 as the first supported version? I'd rather not change the configuration on the connectors side at the moment and adopt this in the 8.6-alpha1.

jonathanlukas:
I can relate to this. But, the new properties structure should still work for old properties.

Is there any way to find this new properties structure in the camunda sdk?

chillleader (approved):
You're right, seems backward compatible. Then I'm fine with backporting it 👍
 
Regarding having this in the upcoming Camunda SDK, I'm not sure how this fits in, since I don't know what kind of structure is planned for that project

PR #741, opened 2024-03-29
chore: prepare bump zeebe version to 8.5.0

houssain-barouni:
Hello,

I wanted to use spring-zeebe with Zeebe 8.5.0 that have now a REST API

I will set zeebe version with 8.5.0 once it is released this week

### DIFF
--- a/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfiguration.java
+++ b/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfiguration.java
@@ -74,38 +74,38 @@ public class ZeebeClientConfiguration implements io.camunda.zeebe.client.ZeebeCl
 
   @Override
   public URI getRestAddress() {
-    return URI.create(camundaClientProperties.getZeebe().getRestAddress().toString());
+    return URI.create(camundaClientProperties.getZeebe().getBaseUrl().toString());
   }
 
   @Override
   public URI getGrpcAddress() {
-    return URI.create(camundaClientProperties.getZeebe().getBaseUrl().toString());
+    return URI.create(camundaClientProperties.getZeebe().getGatewayUrl().toString());
   }
 
   private String composeGatewayAddress() {
     // check if port is set
-    if (camundaClientProperties.getZeebe().getBaseUrl().getPort() != -1) {
+    if (camundaClientProperties.getZeebe().getGatewayUrl().getPort() != -1) {
       String gatewayAddress =
-          camundaClientProperties.getZeebe().getBaseUrl().getHost()
+          camundaClientProperties.getZeebe().getGatewayUrl().getHost()
               + ":"
-              + camundaClientProperties.getZeebe().getBaseUrl().getPort();
+              + camundaClientProperties.getZeebe().getGatewayUrl().getPort();
       LOG.debug("Gateway port is set, address will be '{}'", gatewayAddress);
       return gatewayAddress;
     }
     // check if default port can be applied
-    if (camundaClientProperties.getZeebe().getBaseUrl().getDefaultPort() != -1) {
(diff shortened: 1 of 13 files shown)


jonathanlukas:
@houssain-barouni I would actually map the zeebe rest endpoint to the `base-url` property and use another property `gateway-url` for the gRPC api.

Background: The current configuration is a breaking change and this kind of change can be done. wdyt?

spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfiguration.java
```diff
@@ -350,6 +359,11 @@ public boolean useDefaultRetryPolicy() {
     return false;
   }
 
+  @Override
+  public boolean preferRestOverGrpc() {
+    return false;
```
jonathanlukas, line 364: Make this configurable. Use `camunda.client.zeebe.prefer-rest-over-grpc`.

spring-boot-starter-camunda/src/main/resources/application-camunda-oidc.yaml
```diff
@@ -10,6 +10,7 @@ camunda:
       enabled: true
       base-url: http://localhost:26500
       audience: zeebe-api
+      rest-address: http://localhost:8080
```
jonathanlukas, line 13: This needs to be aligned with the docker-compose from camunda/camunda-platform

spring-boot-starter-camunda/src/main/resources/application-camunda-simple.yaml
```diff
@@ -9,6 +9,7 @@ camunda:
     zeebe:
       enabled: true
       base-url: http://localhost:26500
+      rest-address: http://localhost:8080
```
jonathanlukas, line 12: This needs to be aligned with the docker-compose from camunda/camunda-platform

houssain-barouni:
> @houssain-barouni I would actually map the zeebe rest endpoint to the `base-url` property and use another property `gateway-url` for the gRPC api.
> 
> Background: The current configuration is a breaking change and this kind of change can be done. wdyt?

@jonathanlukas , yes it makes sense. Is `gateway-url` clear enough to show that it is the gRPC address ?

spring-boot-starter-camunda/src/main/resources/application-camunda-saas.yaml
```diff
@@ -9,6 +9,7 @@ camunda:
       enabled: true
       audience: zeebe.camunda.io
       base-url: https://${camunda.client.cluster-id}.${camunda.client.region}.zeebe.camunda.io
+      rest-address: https://${camunda.client.cluster-id}.${camunda.client.region}.zeebe.camunda.io
```
jonathanlukas, line 12: This would mean that grpc and rest are served from the same url in saas. Is that true?

houssain-barouni:
@jonathanlukas zeebe-client-java uses `zeebe.client.gateway.rest.address`, so it may be confusing, isn't? https://github.com/camunda/zeebe/blob/main/zeebe/clients/java/src/main/java/io/camunda/zeebe/client/ClientProperties.java#L46

jonathanlukas:
I see. The "new" configuration moves away from the properties alignment completely, so I don't see an issue.

spring-boot-starter-camunda/src/main/resources/application-camunda-oidc.yaml
houssain-barouni: indeed, it is still not defined which port will be used for Zeebe REST in camunda-platform

spring-boot-starter-camunda/src/main/resources/application-camunda-simple.yaml
houssain-barouni: indeed, it is still not defined which port will be used for Zeebe REST in camunda-platform

jonathanlukas (approved):

PR #797, opened 2024-05-23
Fixes #796 Adding provision to have a result processor for workers

### LINKED ISSUE #796: Adding a provision to have custom result handler (enricher/transformer) in JobHandlerInvokingSpringBeans
We want to do additional processing on the incoming and outgoing variables of the worker before they can go into/come from SaaS. 

We have looked into the code and we see that incoming variable handling can be managed by overriding ParameterResolverStrategy. The same provision is not available for return object. We would like to have a Custom result enricher which can transform the variables before being sent to the Camunda SaaS. So a default enricher which returns the same value can be added to the class JobHandlerInvokingSpringBeans to handle it standard and then it can allow us to override the default enricher with Custom enricher to do additional processing.

We have considered to have this processing in the workers but that makes the code looks unnecessarily cluttered and we want to do this processing in the background. We have also considered using aspects but that doesnt look natural and we want the provision embedded in the framework itself without impacting existing functionality.

**Additional context**

hem-verma:
We want to do additional processing on the incoming and outgoing variables of the worker before they can go into/come from SaaS. Adding the provision to have a custome result processor.

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -1,1 +1,1 @@
-distributionUrl=https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.9.6/apache-maven-3.9.6-bin.zip
+distributionUrl=https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.9.7/apache-maven-3.9.7-bin.zip
--- a/README.md
+++ b/README.md
@@ -305,6 +305,8 @@ public class TestMyProcess {
 
 An example test case is [available here](https://github.com/camunda-community-hub/camunda-cloud-examples/blob/main/twitter-review-java-springboot/src/test/java/org/camunda/community/examples/twitter/TestTwitterProcess.java).
 
+>Please do not use `zeebeTestEngine.waitForBusyState(...)` to wait for a timer. This will not work as this is also triggered by an incoming job activation.
+
 ## Run Connectors
 
 Spring Zeebe project previously included the Runtime for Camunda 8 Connectors. It has been moved to a separate [Connectors](https://github.com/camunda/connectors) project.
--- a/camunda-sdk-java/java-common/pom.xml
+++ b/camunda-sdk-java/java-common/pom.xml
@@ -27,7 +27,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>33.2.0-jre</version>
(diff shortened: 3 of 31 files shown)


jonathanlukas (changes requested):
Thank you for your PR.

I really like the proposed change. However, I would like to reiterate on some details:

* `ResultEnricher` is a opinionated term, I would prefer `ResultProcessor`
* I would split up the `ResultProcessor` into 2 parts (this approach is equal to the `ParameterResolver`):
  * the actual `ResultProcessor` that is able to map the result
  * a `ResultProcessorStrategy` which allows for finding the right `ResultProcessor` for a `@JobWorker` function based on the available information 
* please add unit tests to verify the desired behaviour

hem-verma:
As seen in the screenshot, I have already agreed to the CLA but the same is not reflecting here. Not sure why. Any idea?

image

jonathanlukas:
I think @xomiamoore mentioned that the email address being used for the commit in the PR is different from the one being used to sign the CLA. Could you please verify this?

hem-verma:
That is weird. Seems like I have two accounts or maybe I had an account but later used google account to login again. But the account which created the PR should take the priority. Right? Any other account should be able to push changes as a contributor to the PR. Do you know how this can be fixed?

jonathanlukas:
@xomiamoore provided this guide: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

hem-verma:
Thanks I will check. I have the changes ready as suggested by you. But I have question about the Unit test. Could you please indicate/guide how to write that. I am confused because there is no unit test for JobHandlerInvokingSpringBeans. and there are not many tests available in the zeebe-client module.

jonathanlukas:
The unit test should be able to prove that your configuration is applied so that the beans that depend on the new feature are created correctly.

Also, customization options (like bean overriding) could be tested.

hem-verma:
May I ask if there are existing examples in the repo for testing the implementation for ParameterResolverStrategy? It will help me in understanding how much to test for ResultProcessor. I could not find anything in the folder src/test/java/io/camunda/zeebe/spring/client

hem-verma:
Close this PR because of wrong account (mistakenly) being used for commit. Created a new PR with correct account. https://github.com/camunda-community-hub/spring-zeebe/pull/833

jonathanlukas (commented):
lgtm

PR #804, opened 2024-05-31
bug(object-mapper): Fix our custom mappers creation to avoid mutating the initial ObjectMapper

johnBgood:
## Description

We can't set a process variable to null, because the ObjectMapper was configured to ignore NULL values.
This was a side effect due to the `ZeebeObjectMapper` and `SdkObjectMapper` using the `objectMapper` and mutating it.

To avoid breaking things, I just copied the initial mapper to keep the current behavior for the said mappers while avoiding the mutations.

I plan to backport this in: 
- 8.4
- 8.5

closes https://github.com/camunda/connectors/issues/2648


jonathanlukas (approved):
lgtm

PR #812, opened 2024-06-04
Document timer testing

### LINKED ISSUE #659: zeebeTestEngine.increaseTime not working reliably for Spring Boot Tests
**Describe the bug**

I have a process test whose process includes an timer intermediate catch event with a duration of 15 minutes. Obviously, I don’t want the test to wait but want be be able to trigger the timer (or rather increase the time of the environment so that it fires).

Here is a nice working example which works: [TimerTest.java 1](https://github.com/camunda-community-hub/camunda-8-examples/blob/929e8b6f88e69fd30df5a24650b4d9478a3bb29b/timer-testing/src/test/java/com/camunda/consulting/TimerTest.java)

However, this previous example uses `@ZeebeProcessTest` and in my case I have
`@SpringBootTest @ZeebeSpringTest` and it doesn’t work reliably.

To reproduce the issue I created a fork of the Camunda 8 Twitter Example: [twitter-with-timer 2](https://github.com/tobiasschaefer/camunda-8-examples/tree/twitter-with-timer) in which I

* added a timer with a duration in the process model
* added the same skipTimer implementation.
* disabled all tests but the first (to make it simpler)
* I replaced `@Test` with `@RepeatedTest(10)` to make the error more obvious and reproducible.

However, this test
* sometimes fails on the first zeebeTestEngine.waitForIdleState(DEFAULT)
* sometimes on the last zeebeTestEngine.waitForIdleState(DEFAULT)
* sometimes it succeeds.

See also discussion in the [forum thread #50258](https://forum.camunda.io/t/zeebetestengine-increasetime-not-working-reliably-for-spring-boot-tests/50258/4)

**Expected behavior**

Test should pass every time.

**Log/Stacktrace**

image

Repetition 1: failed on “zeebeTestEngine.waitForIdleState(DEFAULT);” (first line in method skipTimeout) with a timeout

Repetition 2: failed a few lines later on
“zeebeTestEngine.waitForIdleState(DEFAULT);” (last (!) line in method skipTimeout) with a timeout

Repetition 4 failed with “Mockito.verify(twitterService).tweet(“Hello world”);”

On the second invocation all 10 were successful:

image

On further invocations I get random results similar to the first and second invocation.

Note that the duration is quite different. I don’t know much about Zeebe internals but is there some configuration which can be set to make the timers being checked more often?

**Environment:**
- OS: Windows
- Deployment: none - its a local test with `@ZeebeSpringTest`, i.e. it uses the embedded process engine
- Spring-Zeebe Version: 8.3.1
- SpringBoot Version: 3.1.5
- Configuration: n.A.

jonathanlukas:
closes #659

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,8 @@ public class TestMyProcess {
 
 An example test case is [available here](https://github.com/camunda-community-hub/camunda-cloud-examples/blob/main/twitter-review-java-springboot/src/test/java/org/camunda/community/examples/twitter/TestTwitterProcess.java).
 
+>Please do not use `zeebeTestEngine.waitForBusyState(...)` to wait for a timer. This will not work as this is also triggered by an incoming job activation.
+
 ## Run Connectors
 
 Spring Zeebe project previously included the Runtime for Camunda 8 Connectors. It has been moved to a separate [Connectors](https://github.com/camunda/connectors) project.


tobiasschaefer (approved):
Looks good to me.

PR #833, opened 2024-06-10
Fixes #796 Adding provision to have a result processor for workers

### LINKED ISSUE #796: Adding a provision to have custom result handler (enricher/transformer) in JobHandlerInvokingSpringBeans

hem-verma:
We want to do additional processing on the incoming and outgoing variables of the worker before they can go into/come from SaaS. Adding the provision to have a custom result processor.

This PR also contains review comments fixes from previous closed PR. I closed the other one because of the wrong email address being used.

### DIFF
--- a/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientAllAutoConfiguration.java
+++ b/spring-boot-starter-camunda/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientAllAutoConfiguration.java
@@ -1,8 +1,5 @@
 package io.camunda.zeebe.spring.client.configuration;
 
-import static io.camunda.zeebe.spring.client.configuration.PropertyUtil.*;
-import static io.camunda.zeebe.spring.client.properties.ZeebeClientConfigurationProperties.*;
-
 import io.camunda.zeebe.client.api.JsonMapper;
 import io.camunda.zeebe.client.api.worker.BackoffSupplier;
 import io.camunda.zeebe.client.impl.worker.ExponentialBackoffBuilderImpl;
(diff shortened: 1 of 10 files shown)


hem-verma:
Hi @jonathanlukas. Could you please check this one. I closed the other PR (https://github.com/camunda-community-hub/spring-zeebe/pull/797) because I mistakely used a wrong account to push the code. This PR also contains the fixes for your review comments.

hem-verma:
Jonathan's review comments from the other PR.

_Thank you for your PR.

I really like the proposed change. However, I would like to reiterate on some details:

- ResultEnricher is a opinionated term, I would prefer ResultProcessor
-  I would split up the ResultProcessor into 2 parts (this approach is equal to the ParameterResolver): 
- -the actual ResultProcessor that is able to map the result
- - a ResultProcessorStrategy which allows for finding the right ResultProcessor for a @JobWorker function based on the available information
- please add unit tests to verify the desired behaviour_

jonathanlukas (approved):
lgtm

jonathanlukas:
@hem-verma there are formatting issues with the PR. Please make sure to run `mvn clean package` locally and push the resulting changes to be aligned the formatting

hem-verma:
One question. When the release version will be created with this change which we can use?

hem-verma:
I see that the feature is already closed. :)

jonathanlukas:
@hem-verma this feature is released with `8.6` of the client.

Also, this feature will be proposed to be added to the official SDK to be used there as well.

hem-verma:
Thanks @jonathanlukas. Good to know about the official SDK because I also asked the same question to Sean recently.

PR #860, opened 2024-07-01
feat(simple-authentication): add the management of CSRF token for the…

mathias-vandaele:
Add the management of CSRF token for the simple authentication

Another header has to be managed when operate has `CAMUNDA_OPERATE_CSRFPREVENTIONENABLED=true`
This PR allow `spring-zeebe` to manage the CSRF token if it has been enabled.

### DIFF
--- a/camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/Authentication.java
+++ b/camunda-sdk-java/java-common/src/main/java/io/camunda/common/auth/Authentication.java
@@ -4,7 +4,7 @@ import java.util.Map;
 
 public interface Authentication {
 
-  Map.Entry<String, String> getTokenHeader(Product product);
+  Map<String, String> getTokenHeader(Product product);
 
   void resetToken(Product product);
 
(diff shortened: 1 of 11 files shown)


jonathanlukas (commented):
Looks good so far. Did you try this as integration test?

mathias-vandaele:
> Looks good so far. Did you try this as integration test?

I ran all UT, and started the camunda-core docker suite with `CAMUNDA_OPERATE_CSRFPREVENTIONENABLED=true` and `CAMUNDA_OPERATE_CSRFPREVENTIONENABLED=false` and verified with `ExampleApplication` that it was always pulling  tasks correctly from operate

jonathanlukas:
Perfect :) then we can add this here.

Next steps: 
* backport to 8.5
* create issue for dedicated operate client repo

johnBgood:
We'll also need to release `spring-zeebe` and bump its version in `connectors`, we can have a look together @mathias-vandaele if you're not sure how to do it

jonathanlukas:
@johnBgood , after backporting you can simply create a release from the according branch with the next patch version tag which will trigger the action to release the artefacts

johnBgood:
> @johnBgood , after backporting you can simply create a release from the according branch with the next patch version tag which will trigger the action to release the artefacts

Absolutely but I was not sure Mathias knew how to do it :)

PR #943, opened 2024-09-19
Underlying HTTP client evaluates system properties, e.g. proxy settings

tobiasschaefer:
Currently, the underlying HTTP clients cannot be configured to use a proxy (which is unfortunately needed quite of the enterprise environments)

Therefore I suggest this PR to support e.g. http_proxy and https_proxy


tobiasschaefer:
Hi @jonathanlukas ,

could you please have a look at my PR.

I'm tagging you because lately you were most active in this project ;-)

PR #976, opened 2024-10-05
Simplify class assignability check

fml2:


fml2:
I did this in the wrong project!  Did it now in the right one: https://github.com/camunda/camunda/pull/23172

